Issue #6, opened 2017-05-28
Fix CI builds for stable Numpy and Astropy


cdeil:
@bsipocz - I set up this new package `hips` using the package-template cookie-cutter.

One error I'm gettting is this:
```
PackageNotFoundError: Package not found: '' Packages missing in current linux-64 channels: 
  - numpy stable*
  - astropy stable*

You can search for packages on anaconda.org with

    anaconda search -t conda numpy

(and similarly for the other packages)

You may need to install the anaconda-client command line client with

    conda install anaconda-client

Note that you have pinned specs in /home/travis/miniconda/envs/test/conda-meta/pinned:

    ['numpy stable*', 'astropy stable*', 'healpy', 'reproject', 'matplotlib', 'sphinx  1.5.6*']
```
https://travis-ci.org/hipspy/hips/jobs/236899940#L483

Did I mess up the `.travis.yml` setup somehow?
Or is this a bug in the package-template?
Do you know what the fix is or should I investigate?

bsipocz:
This is something weird, I'll try to have a look.

bsipocz:
I have a branch with a cleaned up travis and appveyor file, I'll open a PR once everything is fixed upstream.
First of all, neither ``astropy`` nor ``numpy`` should be listed in ``CONDA_DEPENDENCIES``. It's a known issue in ci-helpers (https://github.com/astropy/ci-helpers/issues/84), and at some point it may be get fixed. 

The other issues is about changes I added to ci-helpers today (when a package is not available on conda we try to install it with pip. there are still some bugs with a few corner cases you hit there, so I'm trying to iron them out before opening a PR here).

cdeil:
@bsipocz - You're https://www.youtube.com/watch?v=mNU3aIJs88g . Thanks!

cdeil:
This issue was fixed by @bsipocz in #9. Closing now.

Issue #3, opened 2017-05-26
Set up Windows testing on Appveyor


cdeil:
Reminder issue for myself to set up windows testing on Appveyor.

cdeil:
A windows build on Appveyor is now running for this project.

It starts failing here:
https://ci.appveyor.com/project/cdeil/hips/build/1.0.4#L86
and then healpy isn't installed by conda and pip install healpy fails.

@bsipocz -- Again the question - is this an issue I introduced in `appveyor.yml` or an issue with the `package-template`? Can you help and make it better?

bsipocz:
There is no conda build for healpy and numpy 1.13. This case ci-helpers tries to install the packages with pip, but I simply guess that is not possible for healpy on windows? I'll change the numpy version in a joint PR with #6.

cdeil:
This is done. Windows build on Appveyor is working since / thanks to the switch to astropy-healpix: #109 . I'll make a v0.2 release now to get this change / added platform support available for users.

PR #9, opened 2017-05-30
Travis cleanup

bsipocz:
This should ultimately address both #3 and #6, but there are still upstream issues. Opening it now as a WIP, so less likely it slips out of my mind.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,6 @@ os:
 # Setting sudo to false opts in to Travis-CI container-based builds.
 sudo: false
 
-# The apt packages below are needed for sphinx builds. A full list of packages
-# that can be included can be found here:
-#
-# https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise
-
 addons:
     apt:
         packages:
@@ -23,7 +18,6 @@ addons:
 
 env:
     global:
-
         # The following versions are the 'default' for tests, unless
         # overridden underneath. They are defined here in order to save having
         # to repeat them for all configurations.
@@ -41,10 +35,9 @@ env:
 
     matrix:
         # Make sure that egg_info works without dependencies
-        - PYTHON_VERSION=3.5 SETUP_CMD='egg_info'
+        - SETUP_CMD='egg_info'
 
 matrix:
-
     # Don't wait for allowed failures
     fast_finish: true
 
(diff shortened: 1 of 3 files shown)


bsipocz:
Btw a nice workaround for the pip install issues would be to update the conda-forge build for healpy to get the latest version (rather than 1.9.1). That case it would be built with newer numpies...

bsipocz:
Actually I have no idea why the egg_build disappears. It does the same thing when I build the latest hips master on my fork, too (https://travis-ci.org/bsipocz/hips/builds/237675420)

cdeil:
@bsipocz - Can you please leave the Appveyor build in place? At the moment we don't import healpy yet, so it should be possible.

I think we definitely want to support Windows here, this is a package that many hobby astronomers or schools might like: an easy way to fetch images and catalogs for any part of the sky, and expose it in a form so that it can be analysed with Python.

The functionality we need from healpy here is very limited, and if they don't start supporting Windows soon, I might actually take the time and implement https://github.com/cdeil/healpix and then we use that here instead of healpy.

bsipocz:
OK, fair enough. To be honest I only checked the installation docs to see whether it's a requirement or an optional one.

bsipocz:
(close-reopen: just pushed a fix to ``ci-helpers``, and needed to retrigger travis on this PR)

bsipocz:
@cdeil - This should be good to go for travis, there's one docs warning left only, I haven't paid much attention to appveyor. Also you may want to opt in the auto cancellation on travis, so outdated commits get cancelled out of the queue.

cdeil:
@bsipocz - https://travis-ci.org/hipspy/hips/jobs/238045310#L1023 ?

For https://travis-ci.org/hipspy/hips/jobs/238045306#L1083 , can you just add this to the toc in `index.rst` or should I do that in master?

cdeil:
I didn't know about travis-ci auto-cancel. Thanks!
I've activated it for this `hips` repo and `gammapy` as well.

bsipocz:
@cdeil - Yes, they have it for about 6 months now, and just moved it out from beta testing this week. In astropy we use it since early beta times, works quite well.

bsipocz:
Also I can do the toc fix tomorrow.

cdeil:
> also I guess you don't want to deal with astropy 1.0.x compatibility?

No. Even just supporting Astropy 1.3 or later for this package would be fine.

bsipocz:
@cdeil - This should be ready now and get a green travis.

cdeil:
🎉

PR #11, opened 2017-05-31
Add skeleton class TileFileCache

adl1995:
This class will provide functionality such as reading data from the `hips-extra` repository and
fetching tiles from remote servers (like CDS) into the local tile cache.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,7 +4,7 @@
 Python HiPS package documentation
 #################################
 
-.. attention::
+.. note::
 
     This package is being developed as a Google Summer of Code 2017 project. The progress 
     of this package will be updated on this `blog <https://adl1995.github.io>`_ and the student's
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,7 +4,7 @@
 Python HiPS package documentation
 #################################
 
-.. note::
+.. attention::
 
     This package is being developed as a Google Summer of Code 2017 project. The progress 
     of this package will be updated on this `blog <https://adl1995.github.io>`_ and the student's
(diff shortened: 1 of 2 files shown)


cdeil:
@adl1995 - There's some boilerplate involved. Please do for `hips.tiles` the equivalent of what I did for `hips.draw` here:
https://github.com/hipspy/hips/commit/4fc6601f464edef9c6f85eb71349dfec46cf064b

One check you can do is
```
python setup.py build_docs
open docs/_build/html/index.html
```
and check if your new class appears there.

For this to work, you need to add the `__all__`, add the `__init__.py` files with the imports, like I did for `hips.tiles`.
Let me know if you run into any issues.

----

Apart from that, I would always suggest to start with a high-level docs example or test that exercises the new functionality for the "common case" and "user" side. This will help us develop a package / an API that's user-friendly, and to focus on developing the key functionality first, instead of getting side-tracked to develop "infrastructure" like cache stuff that we think we'll need, but then at the end don't really need.

When you add classes and methods and high-level tests like that, at first they will not work, because they aren't implemented yet. What you can do is put `raise NotImplementedError` into the methods, and put a `@pytest.mark.xfail` on the tests. Then once we've discussed the code structure and API a bit, you would start to implement it and make it work piece by piece.

OK?

cdeil:
@adl1995 - What's the status with this pull request?
It's a dummy class without anything implemented / no tests or working code.

Do you want to continue and implement something here, or close this and start / continue with some other part (e.g. #24 or other things we talked about in the last 2 weeks)?

cdeil:
@adl1995 - Close this PR and continue coding in #33 ?
(or the other way around: your choice)

adl1995:
Closing this PR. This will instead be implemented in #33.

PR #12, opened 2017-06-01
Sphinx update

adl1995:
Added link to my GSoC application, blog, and updated the note directive.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,7 +4,7 @@
 Python HiPS package documentation
 #################################
 
-.. attention::
+.. note::
 
     This package is being developed as a Google Summer of Code 2017 project. The progress 
     of this package will be updated on this `blog <https://adl1995.github.io>`_ and the student's
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,7 +4,7 @@
 Python HiPS package documentation
 #################################
 
-.. note::
+.. attention::
 
     This package is being developed as a Google Summer of Code 2017 project. The progress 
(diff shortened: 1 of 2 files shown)


cdeil:
@adl1995 - Thanks for splitting this docs note update out into a separate pull request.

When you make a pull request, please always review the diff and commmits (yourself, before opening the PR and Thomas or I start reviewing it).

In this case, looking at https://github.com/hipspy/hips/pull/12/files, I still see `hips/tiles/tilefilecache.py` from #11 which shouldn't be here. Also, there's 10 commits in https://github.com/hipspy/hips/pull/12/commits for something that should be a single-commit, single-addition in one file (so that we get a sensible / clean git history for this repo).

Probably the root of these git issues is that you started #12 in `adl1995:master`, i.e. started to do work and commits in your local `master` branch. You should never do that, but always start the work by making a new feature branch. See e.g. http://astropy.readthedocs.io/en/latest/development/workflow/development_workflow.html#workflow

Please take a few hours to read through some git / github info (e.g. a tutorial or Github docs, or the Astropy dev guide) and let me know if you have any question.

Now how to clean up? One simple option could be to close #11 and this pull request, and to start fresh with new branches / pull requests. Of course, with git, it's also possible to clean up / recover with your existing branches / pull requests. But that requires some more knowledge of git like for example using `git rebase` or `git reset`. If you want, you could try and practice git for this `sphinx-update` branch, and get rid of `hips/tiles/tilefilecache.py` and squash all the commits into a single one with the edit in `docs/index.rst`.

PR #15, opened 2017-06-02
Small docs update

cdeil:
This PR:
- Changes the table of content and sidebar as shown in the screenshot
- Moves the "tile drawing algorithm page" into the developer docs section (as most users won't read through it) and shortens the headings so that they are readable in the table of content side bar. Otherwise it's just whitespace formatting changes. Hope that change is OK @tboch .
- Adds a very small page with developer info (we can extend later)

screen shot 2017-06-02 at 22 36 45

### DIFF
--- a/docs/develop.rst
+++ b/docs/develop.rst
@@ -1,0 +1,19 @@
+.. include:: references.txt
+
+.. _develop:
+
+*******
+Develop
+*******
+
+Want to contribute to the ``hips`` package?
+
+Great! Talk to us by filing a Github issue any time
+(it doesn't have to be a concrete feature request or bug report).
+
+This package was created using the Astropy affiliated package template,
+and everything works pretty much as in Astropy and most affiliated packages.
+
+We didn't write any developer docs specifically for this package yet.
+For now, check out the Astropy core package developer docs,
+or just talk to us if you have any questions.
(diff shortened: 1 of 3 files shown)


cdeil:
Merging this now. Comments / suggestions still welcome!

tboch:
this update is fine for me

adl1995:
This looks good! Should we also add a section named "Guides" as we move along. This could contain small tutorial(s) on how to use the package.

cdeil:
>  Should we also add a section named "Guides" as we move along? This could contain small tutorial(s) on how to use the package.

My suggestion would be that as we add functionality, we make sure to add good docstrings (in some cases with `Examples` sections at the end, so that over time our API docs become nice:
https://hips.readthedocs.io/en/latest/api.html

For high-level, tutorial-style docs, as we develop functionality, we fill this page:
https://hips.readthedocs.io/en/latest/getting_started.html

And then if we have developed a lot of functionality and the single getting started page gets too long, we have a conversation about our docs and refactor it into a larger set of pages.

That said, @adl1995 - of course suggestions / pull requests are welcome any time!
Just in case it's something larger that you'd have in mind (i.e. that would take a day or more to write up), please make an issue proposing / outlining the addition or change before implementing it.

PR #20, opened 2017-06-05
Update hips/conftest.py

cdeil:
This pull request causes the test runner to print out more useful info at the top, specifically:
- the hips version number
- the healpy version number

@adl1995 - I'm adding this because in this build
https://travis-ci.org/hipspy/hips/jobs/239317653#L1006
from your PR #19 it's not clear from the log which healpy version is used there.


cdeil:
Now when running the tests, the hips version is printed:
https://travis-ci.org/hipspy/hips/jobs/239753329#L1026
and the healpy version is printed:
https://travis-ci.org/hipspy/hips/jobs/239753329#L1048

Merging this now.

PR #21, opened 2017-06-09
Add tiles.hipsdescription module

adl1995:
This commits adds a class named HipsDescription which provides methods for parsing the HiPS properties. It parses the given string and returns an object of OrderedDict. Additional getter methods are also included.
For the test cases, a sample properties file was obtained from http://alasky.unistra.fr/DSS/DSSColor/properties


docs/api.rst
```diff
@@ -2,3 +2,4 @@ Reference/API
 =============
 
 .. automodapi:: hips
+.. automodapi:: hips.tiles
```
cdeil: I would suggest that, at least for now, we import everything (except the utils) into the top-level `hips` namespace (like Numpy does it) instead of splitting things in sub-packages (like Scipy does it). The sub-packages and sub-files are only to organise the code into multiple files, but users then import everything from the `hips` namespace.
So instead of adding the automodapi directive here, I'd suggest you add a line `from .tiles import *` here:
https://github.com/hipspy/hips/blob/master/hips/__init__.py#L21
Also, there is now a merge conflict for `docs/api.rst`. Please rebase this PR against latest master.

hips/tiles/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""HiPS description class -- provides methods for parsing the HiPS file properties."""
```
cdeil: This is not the appropriate description of the `hips.tiles` package. It will contain many classes.
Suggest to remove this docstring here, or to replace it with something like
> Classes and functions to manage HiPS tiles.

hips/tiles/hipsdescription.py
```diff
@@ -0,0 +1,75 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""HiPS description class -- provides methods for parsing the HiPS file properties."""
+
+__all__ = [
+    'HipsDescription',
+]
+
+from collections import OrderedDict
```
cdeil: Usually the `__all__` list comes below the imports. Please change the order in this file.

hips/tiles/hipsdescription.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class HipsDescription:
+    """This class provides methods for parsing the HiPS properties file.
```
cdeil: Docstrings should always be a summary line, followed by an empty line, followed by more description or sections like `Parameters` etc.
The pattern we use in Astropy is to document the parameters of `__init__` in the `Parameters` section of the class-level docstring (and to never put a docstring on `__init__`. So here, please add it here to describe the `properties` parameter.

hips/tiles/hipsdescription.py
```diff
@@ -0,0 +1,75 @@
...
+        ----------
+        properties : str
+            HiPS file properties
+        
+        Returns
+        -------
+        dict_properties : OrderedDict
+        """
+        properties = properties.split('\n')
+        dict_properties = {}
```
cdeil: If you first create a regular dict, you've lost the order of the input.
You should start with an empty `OrderedDict` and then add entries.
Or create a list of `(key, value)` tuples and pass that list to the `OrderedList` constructor, that will also preserve the order.

hips/tiles/hipsdescription.py
```diff
@@ -0,0 +1,75 @@
...
+        for property in properties:
+            key_value = property.split('=')
+            try:
+                dict_properties[key_value[0].strip()] = key_value[1].strip()
+            except:
```
cdeil: Why is this `try-except` needed at all?
If it's not needed, please remove it.
If it's needed, catch the specific exception that occurs and leave a one-line comment why / when it occurs (e.g. by giving an example of a line where this happens.

hips/tiles/hipsdescription.py
```diff
@@ -0,0 +1,75 @@
...
+                pass
+        return OrderedDict(dict_properties)
+
+    def get_properties(self) -> OrderedDict:
```
cdeil: This isn't useful, this method can be removed.
In Python, it's considered OK / normal to just have public attributes on classes by default, and to not have any getter / setter methods. Only if needed do you "lock down" classes and try to hide the data members and use getter / setters.

hips/tiles/hipsdescription.py
```diff
@@ -0,0 +1,75 @@
...
+        """OrderedDict: Returns the ordered dictionary containing the HiPS properties."""
+        return self.properties
+
+    def get_base_url(self) -> str:
```
cdeil: I think you should change all of these
```
def get_something(self):
    return some_type(self.properties['something'])
```
to
```
@property
def something(self):
    return some_type(self.properties['something'])
```
i.e. use Python properties instead of getter methods, when there are no arguments to the getter.

Also, is it useful  to try-except and return none if the info isn't there?
If you don't have the try-except you'll get a KeyError and that's a bit better no?
I mean: if a caller doesn't handle the case of a non-existent key, then getting an error at that point is better than getting none and failing later.
and if a caller can handle the case of a non-existent key, she can just as easy use try-except KeyError and decide what to put, as it would be to decide what to do with the None return value.

hips/tiles/hipsdescription.py
```diff
@@ -0,0 +1,75 @@
...
+    def get_title(self) -> str:
+        """str: Returns the title from the HiPS file properties."""
+        return self.properties['obs_title']
+
+    def get_hips_version(self) -> float:
+        """float: Returns the HiPS version from the HiPS file properties."""
+        return float(self.properties['hips_version'])
+
+    def get_hips_frame(self) -> str:
+        """str: Returns the HiPS frame from the HiPS file properties."""
```
cdeil: The common one-line docstring pattern we have for properties is to mention the type at the end of the line in parentheses, like this:
```
@property
def something(self):
     """Something (`some_type`)."""
```

hips/tiles/tests/test_hipsdescription.py
```diff
@@ -0,0 +1,79 @@
...
+        assert cls.hipsdescription.get_hips_version() == 1.31
+
+    def test_get_hips_frame(cls):
+        assert cls.hipsdescription.get_hips_frame() == 'equatorial'
+
+    def test_get_hips_order(cls):
+        assert cls.hipsdescription.get_hips_order() == 9
+
+    def test_get_tile_format(cls):
+        assert cls.hipsdescription.get_tile_format() == 'jpeg'
```
cdeil: Add newline at end of file.

hips/tiles/tests/test_hipsdescription.py
```diff
@@ -0,0 +1,79 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+from .. hipsdescription import HipsDescription
```
cdeil: Remove the space between `..` and `hipsdesription`.

cdeil:
@adl1995 - Can you please wrap up this pull request?

Once you've implemented comments, always leave a comment asking me to have another look, or if you don't know how to address one of my comments, leave a comment asking for clarification. If you just push new commits, I usually won't look at a PR until you ping me.

In this case: you're still not taking the `OrderedDict` as input in `__init__`, like I suggested in an inline comment above. Can you please change this? Is it clear what to do?

hips/tiles/description.py
```diff
@@ -0,0 +1,99 @@
...
+    It also provides multiple getters around the commonly used HiPS properties.
+
+    Parameters
+    ----------
+    properties : str
+        A string containing HiPS tile properties
+    """
+
+    def __init__(self, properties: str) -> None:
+        super(HipsDescription, self).__init__()
```
cdeil: `HipsDescription` has no super-class (other than the default `object` in Python). No need to call `super` here, you can remove this line.

hips/tiles/description.py
```diff
@@ -0,0 +1,99 @@
...
+        Parameters
+        ----------
+        filename : str
+            Name of the file containing HiPS file properties
+
+        Returns
+        -------
+        dict_properties : OrderedDict
+        """
+        file = open(filename, 'r')
```
cdeil: Always use this pattern to read (and write) files (to make sure they are closed immediately after the reading is done, in a predictable manner:
```
with open(filename) as fh:
    text = fh.read()
return cls.parse(text)
```

adl1995:
@cdeil I just made another commit. Let me know if the `__init__` method is being called in the correct manner now.

cdeil:
@adl1995 - I've just now pushed a commit to your branch for this PR: 3f9faed233a0d10cc771532d09f79eff7d279450

Some changes:
- Shorter name for the classmethods
- Fix return type annotation for classmethods
- Change version property to string (float is no good, because floats often can't be represented exactly on computers)
- Move example data file into a separate text file so that it's easier to read the test code file, and tha the `HipsDesription.read` method can be tested (and adding the common Astropy boilerplate of `setup_package.py` that makes sure the data file gets installed into site-packages with the package, and using an Astropy helper function to create the filename relative to the text module)
- Shorter docstrings
- Some other minor misc cleanup

I think this is ready to be merged, but I'll wait for travis-ci to run or comments until tomorrow.

cdeil:
@adl1995 - Thanks!

PR #22, opened 2017-06-09
Update docstring

adl1995:
This pull request updates the docstring to show the order of the returned corners. Also, the default value for the `nest` parameter is set to `True`.


cdeil:
Thanks!

PR #24, opened 2017-06-14
Add HipsTile and HipsTileMeta classes

adl1995:
This pull request adds a class called `HiPS` which provides methods to handle the HiPS base url and tile url.


cdeil:
Just `Hips` is not a good name. It's too generic, everything in this package is related to Hips.

The functionality you have here now is basically a function that takes three parameters and builds a URL string:
```
hips.get_tile_url(order=6, ipix=24185, format='jpg') == 'http://alasky.unistra.fr/DSS/DSSColor/Norder6/Dir20000/Npix24185.jpg'
```

@adl1995 - I'm not sure how best to proceed with this PR.

You could either change the class you have here into a simple function `make_hips_tile_url` (or some similar name) and also pass in `base_url` there.

Or you could extend the PR and implement some of the things we discussed via email and in Google hangouts in the past 2 weeks, e.g. a class to represent a tile, and then leave this `get_tile_url` as a method on that class.

One more comment: eventually we will also need to read / write tiles from disk. There you should use [pathlib.Path](https://docs.python.org/3/library/pathlib.html) objects instead of just strings, because the folder separator is `\` on Windows and not `/` like you have here.
So in the end, there will be at least two very much related functions or methods or properties that create a URL or filename (and for the filename case there is no "base url", but possible a "base path". I'm OK with refactoring whatever you add here later, unless you already have an idea how to structure the code to support these two use cases.

adl1995:
@cdeil - I have modified the class name to `HipsTile`.

I also added some extra methods. Should this class also handle a tile header? If so, a tile header would only be available if the format is 'FITS', right? In other cases, should I just set the variables to `None`? I was also thinking of adding additional methods such as constructing a WCS object from a HiPS tile, but then again this would only be possible if the tile format is 'FITS'.

Would it be appropriate if I merged the fetching part within this class or should I add a separate class for that?

I'm also getting this error when I try and run the tests:

```
hips/tiles/tile.py:45: in data
    self.data = data
hips/tiles/tile.py:45: in data
    self.data = data
!!! Recursion detected (same locals & position)
```

But there is no recursion involved! The issue came when I modified the `__init__` method to set the values. I haven't found a solution except for this [issue](https://github.com/pytest-dev/pytest/issues/70) but I have not been able to solve this yet.

cdeil:
@adl1995 - I think you can just pass `data=None` in `__init__` and set is as `self.data = data` there and remove the property getter / setter you have for data.

So if this class represents a tile, the most important functionality is that you're able to read a tile from local disk, or from a URL. E.g. you could start with this tile and try to get it to work in your test:
https://github.com/hipspy/hips-extra/raw/master/datasets/samples/DSS2Red/Norder3/Dir0/Npix448.fits

First, try to get the reading of the pixel values to work (and e.g. assert on the pixel `data[10, 20]` value in the test). Tile metadata handling will be a topic, but to just draw tiles, what you need is the pixel values, no?

hips/tiles/tile.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def data(self) -> list:
+        """HiPS tile data (`list`)."""
+        return self.data
+
+    @property
+    def tile_url(self) -> str:
+        """HiPS tile url (`str`)."""
+        directory = np.around(self.ipix, decimals=-(len(str(self.ipix)) - 1))
```
cdeil: Split out this line computing `directory` into a separate "private" method:
```
@staticmethod
def _directory(ipix):
    """Directory of the tile (int)"""
    return np.around(ipix, decimals=-(len(str(ipix)) - 1))
```
The advantage is that then it's easier to test and (if needed, possibly) can be called from other places than `tile_url` (e.g. from a `tile_filename` you'll need later to access tiles from local disk).

hips/tiles/tile.py
```diff
@@ -0,0 +1,45 @@
...
+import numpy as np
+
+__all__ = [
+    'HipsTile',
+]
+
+
+class HipsTile:
+    """HiPS tile container.
+
```
cdeil: Document `Parameters` here, and also add an `Examples` section, showing how to fetch a tile and how to access it's pixel data.

cdeil:
One more comment: I'd suggest to use here (and always) the pattern where `__init__` just takes the data members and stores them away on `self`, and things like file I/O or fetching tiles are done in `classmethods` like `read`, i.e. the pattern we already used for the one class we currently have: https://hips.readthedocs.io/en/latest/api/hips.HipsDescription.html#hips.HipsDescription.read

adl1995:
@cdeil I made a commit which adds some extra methods for reading tiles (from disk and URL). It also provides a method for storing a HiPS tile on disk. The default read / write path is `hips/tiles/tests/data`. These are all standard class methods. As I wanted access to instance variables, I could not use `@classmethd`. The docstring is updated to contain an example showing fetching of a HiPS tile.

I'm still not able to test the `fetch` method as it tries to access data from a remote url. Is there any workaround for this?

hips/tiles/tests/test_tile.py
```diff
@@ -0,0 +1,48 @@
...
+from pathlib import Path
+import numpy as np
+
+class TestHipsTile:
+    @classmethod
+    def setup_class(cls):
+        filename = get_pkg_data_filename('data/properties.txt')
+        hips_description = HipsDescription.read(filename)
+
+        cls.tile = HipsTile(hips_description, 6, 24185, 'jpg', 512)
```
cdeil: I suggest to call this function with keyword arguments, i.e. `order=6`, ... instead of just `6`, to make it easier to read, i.e. know what the parameters are without having to look up the class docs.
If the lines get to long, wrap them (Python is OK with this within parentheses).

hips/tiles/tests/test_tile.py
```diff
@@ -0,0 +1,48 @@
...
+        assert self.tile.base_url == 'http://alasky.unistra.fr/DSS/DSSColor'
+
+    def test_tile_url(self):
+        assert self.tile.tile_url == 'http://alasky.unistra.fr/DSS/DSSColor/Norder6/Dir20000/Npix24185.jpg'
+
+    def test_fetch(self):
+        self.tile.fetch()
+
+        """
+        This data was obtain from 'http://alasky.unistra.fr/DSS/DSSColor/Norder6/Dir20000/Npix24185.jpg'
```
cdeil: Remove this comment? (It doesn't give extra info compared to the code)

hips/tiles/tests/test_tile.py
```diff
@@ -0,0 +1,48 @@
...
+        at the index [10, 20]
+        """
+        data_precomp = [0, 0, 0]
```
cdeil: Suggest to change the pixel (and maybe the tile) to have a pixel with actual values, not `rgb=[0, 0, 0]` like you have for this pixel.

hips/tiles/tests/test_tile.py
```diff
@@ -0,0 +1,48 @@
...
+        assert self.data[10, 20] == data_precomp
+
+    def test_read(self):
+        self.tile.read()
+
+        """
+        This data was obtain from 'http://alasky.unistra.fr/DSS/DSSColor/Norder6/Dir20000/Npix24185.jpg'
```
cdeil: Again, suggest to remove this comment.

hips/tiles/tile.py
```diff
@@ -0,0 +1,131 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from astropy.utils.data import get_pkg_data_filename
```
cdeil: I think this `from astropy.utils.data import get_pkg_data_filename` can be removed from this file, it isn't used, no?

hips/tiles/tile.py
```diff
@@ -0,0 +1,131 @@
...
+__all__ = [
+    'HipsTile',
+]
+
+
+class HipsTile:
+    """HiPS tile container.
+
+    This class provides methods for fetching, reading,
+    and stroing a HiPS tile. It also contains a few
```
cdeil: `stroing` -> `writing`

hips/tiles/tile.py
```diff
@@ -0,0 +1,131 @@
...
+        Class HipsDescription contains HiPS properties
+    order : `int`
+        Order of the HiPS
+    ipix : `int`
+        HEALPix pixel number
+    tile_width : `int`
+        Width of HiPS tile
+    format : `format`
+        Format of HiPS tile
+    data : `list`
```
cdeil: `data` should never be stored in a list, always a `numpy.ndarray`.

hips/tiles/tile.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @staticmethod
+    def _directory(ipix: int) -> int:
+        """Directory of the HiPS tile (`int`)."""
+        return np.around(ipix, decimals=-(len(str(ipix)) - 1))
+
+    @property
+    def base_url(self) -> str:
+        """HiPS tile base url (`str`)."""
+        return self._base_url
```
cdeil: There is no `_base_url`, this is a bug, no?

hips/tiles/tile.py
```diff
@@ -0,0 +1,131 @@
...
+        Parameters
+        ----------
+        filename : `str`
+            Name of the file
+        """
+        path = self.path.joinpath(''.join([filename, '.', self.format]))
+        print(path)
+        with path.open('w') as wf:
+            if self.format is 'fits':
+                hdu = fits.PrimaryHDU(self.data, header=self.header)
```
cdeil: There is no `self.header`, i.e. this is broken code, no?
Please when adding code, **always** add tests in the same pull request that execute the code for at least one case to show that it's working.
You can check coverage locally while you're coding with
```
python setup.py test --coverage
open htmlcov/index.html
```

hips/tiles/tile.py
```diff
@@ -0,0 +1,131 @@
...
+    def store(self, filename: str) -> None:
+        """Store HiPS tile by a given filename (`None`).
+
+        Parameters
+        ----------
+        filename : `str`
+            Name of the file
+        """
+        path = self.path.joinpath(''.join([filename, '.', self.format]))
+        print(path)
```
cdeil: Remove print call here.

hips/tiles/tile.py
```diff
@@ -0,0 +1,131 @@
...
+        with path.open('w') as wf:
+            if self.format is 'fits':
```
cdeil: change `is` to `==` here, and read up on the difference between the two (Google / Stackoverflow).

Also, I'd suggest to move the `if` on format to the top level, i.e. have the `with path.open()` inside the if block and only where you need it. E.g. to write FITS, you make a call to `image_hdu.write()` and never have to call `path.open` yourself.

hips/tiles/tile.py
```diff
@@ -0,0 +1,131 @@
...
+        with path.open('rb') as rf:
+            raw_img = BytesIO(rf.read())
+            if self.format is 'fits':
+                hdulist = fits.open(raw_img)
+                self.data = np.array(hdulist[0].data)
+                self.header = hdulist[0].header
+            else:
+                self.data = np.array(Image.open(raw_img))
+
+    def store(self, filename: str) -> None:
```
cdeil: Rename `store` -> `write`.
It's the common naming pattern in Astropy (e.g. `Table.write`) and we should adopt it instead of using `store` for the same thing in some places.

adl1995:
@cdeil I added the `@remote_data` decorator after importing it using: `from astropy.tests.helper import remote_data`, but I think this just skips the test case. I tested this by adding a line `asset 1 == 2` and this did not raise any error.

cdeil:
Did you try 
> Turn on the remote data tests at the command line with python setup.py test --remote-data=any.
as suggested here?
http://docs.astropy.org/en/stable/development/testguide.html#working-with-data-files

hips/tiles/tile.py
adl1995: Actually, this is required for the `HipsDescription` constructor, as it needs the properties file text.

hips/tiles/tile.py
adl1995: I was assigning the `base_url` after initialization through `@base_url.setter`, and accessing it through `@property`. But, I have removed the setter and getter now as this can be done without that as well.

hips/tiles/tile.py
adl1995: I have moved the `if` statement to the outer block, but this requires me to call `with path.open('w') as wf` twice:

```
if self.format == 'fits':
    with path.open('w') as wf:
        hdu = fits.PrimaryHDU(self.data, header=self.header)
        hdulist = fits.HDUList([hdu])
        hdulist.writeto(wf)
        hdulist.close()
else:
    with path.open('w') as wf:
        Image.fromarray(self.data).save(wf)

```

hips/tiles/tile.py
cdeil: No, it's not used in the constructor (only in the docstring example, which is independent).
Try removing this import line and you'll see that everything still works (I'm pretty sure, but not 100%, I didn't try myself).

hips/tiles/tile.py
cdeil: For the case of FITS, you can just call the `writeto` method on the `PrimaryHDU` object, which takes a filename as argument. No need to call `path.open` yourself or to make an `HDUList` before writing (these things are done in the background by the `PrimaryHDU.write` call).

Example:
```python
import numpy as np
from astropy.io import fits
hdu = fits.PrimaryHDU(np.ones((10, 20)))
hdu.writeto('data.fits')
```

hips/tiles/tile.py
adl1995: The `self.header` property is always populated when the file format is `fits`. In this case `write` is called after `read` or `fetch`, so the header is present.

I ran `python setup.py test --coverage`, but I can't find any directory named `htmlcov`.

hips/tiles/tile.py
cdeil: Actually, also for the `PIL.Image` case, you should be able to just pass a filename to the `Image.save` call, no need to call `path.open` yourself and pass a file handle.
See http://pillow.readthedocs.io/en/4.1.x/reference/Image.html#PIL.Image.Image.save

hips/tiles/tile.py
adl1995: The `HipsDescription` constructor only expects the filename, but in order to get that, `get_pkg_data_filename` is required. This is the line that does it:

```
filename = get_pkg_data_filename('data/properties.txt')
hips_description = HipsDescription.read(filename)

```
This is how it was done in [test_description.py](https://github.com/hipspy/hips/blob/master/hips/tiles/tests/test_description.py) as well. I tried running the test by only passing the filename as a string, but the test case fails, giving this error:

> FileNotFoundError: [Errno 2] No such file or directory: 'data/properties.txt'

cdeil: You have this import line in `hips/tiles/tile.py`.
Yes, you need it in tests files.
But no, you don't need it in `hips/tiles/tile.py`.

adl1995:
@cdeil - I just made a commit which addresses the issues you mentioned above.

cdeil:
@adl1995 - Thinking about the `HipsTile` class a bit more and looking at the code, I noticed that there's some things that aren't nice, probably indicating that we haven't found good abstractions yet / should introduce extra classes.

My first point is that in `HipsTile`, you store a [HipsDescription](https://hips.readthedocs.io/en/latest/api/hips.HipsDescription.html) (require it in `__init__`), but then never access it, i.e. don't use it for anything. 

Note that `HipsDescription` is **not** the "meta" or "header" information for a single tile, even if it looks a bit that way. It is the meta or header information for a whole Hips survey, usually consisting of 1000s of tiles at different orders / sky positions.

I think for now `HipsDescription` should simply be removed from the `HipsTile` class, and maybe later we introduce a class separate out the "meta" information (like order / ipix) from `HipsTile` into a `HipsTileMeta` class.
And in addition, I would suggest you rename `HipsDescription` to `HipsSurveyProperties` (or `HipsSurveyMeta` or  ...) and update the docstring so that it's clear that this is for a given survey, this is not the metadata for a single tile. Also, adding and example to the `HipsDescription` docstring that shows how to fetch one properties.txt file and links to it from the docstring (so that the user can quickly look at the example) would be helpful. @adl1995 - Can you make this change for `HipsDescription` please? Probably best in a separate PR.

---

Other than that, we'll have to discuss more closely how read / write / fetch exactly works and which info for that should be `HipsTile` class data members or not. At the moment it's a pretty wild mix, with `base_url` and `format` being data members, but e.g. `base_path` for the filename not. Intuitively I would think that the `HipsTile` class should be minimal and then methods like `fetch` or `read` become classmethods that take a base_url and filename as input (and write takes the full filename as output. Would that work, or would we need to introduce a second class to generate urls and filenames then?

hips/tiles/tile.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @staticmethod
+    def _directory(ipix: int) -> int:
+        """Directory of the HiPS tile (`int`)."""
+        return np.around(ipix, decimals=-(len(str(ipix)) - 1))
+
+    @property
+    def path(self) -> Path:
+        """Default path for tile storage (`Path`)."""
+        return Path('hips', 'tiles', 'tests', 'data')
```
cdeil: This hard-coded `Path('hips', 'tiles', 'tests', 'data')` is weird and not useful. Either this class is responsible for generating paths matching the pattern tiles are stored with Hips, or someone else does it, and simply full filenames are passed as input to read / write method of this class.

hips/tiles/tile.py
```diff
@@ -0,0 +1,124 @@
...
+        self.ipix = ipix
+        self.tile_width = tile_width
+        self.format = format
+        self.data = data
+        self.base_url = base_url
+
+    @staticmethod
+    def _directory(ipix: int) -> int:
+        """Directory of the HiPS tile (`int`)."""
+        return np.around(ipix, decimals=-(len(str(ipix)) - 1))
```
cdeil: It's weird that this integer is called "directory".
@adl1995 - Do you understand what exactly it is? (it's not a HEALPix order, is it?)
Does it ever make sense to have this int or do we only ever use it in combination with the `Norder` part of the string? If so, maybe add that in here and return the folder name as a string?

hips/tiles/tile.py
```diff
@@ -0,0 +1,124 @@
...
+        if self.format == 'fits':
+            hdulist = fits.open(raw_image)
+            self.data = np.array(hdulist[0].data)
+            self.header = hdulist[0].header
+        else:
+            self.data = np.array(Image.open(raw_image))
+
+    def read(self) -> None:
+        """Read HiPS tile data from a directory and load into memory (`None`)."""
+        path = self.path.joinpath(''.join(['Npix', str(self.ipix), '.', self.format]))
```
cdeil: Note that `pathlib.Path` objects have a `/` operator defined, i.e. you can use `path / 'myfolder' /  'myfile.png' in combination with strings. Please use that pattern everywhere instead of calling `joinpath`

hips/tiles/tile.py
adl1995: This is used in combination with `/Dir`. I'll concatenate them both and return them as a string.

hips/tiles/tile.py
adl1995: This could be moved to `utils.py` as a simple function which returns path for tile storage. Again, this means that the `fetch`, `read`, and `write` would have to be modified to take the updated parameters.

adl1995:
@cdeil - My latest commit removes most of the functionality from the class. I will move the previous method from the class to `utils.py` when [this](https://github.com/hipspy/hips/pull/26) PR gets merged, otherwise I think there will be Git conflicts. For now, there is a `@property` in this class, but I'm thinking of moving it to `utils.py` as well under the name `base_path`, where all the HiPS tiles will be read / written to.

hips/tiles/tile.py
```diff
@@ -0,0 +1,114 @@
...
+class HipsTile:
+    """HiPS tile container.
+
+    This class provides methods for fetching, reading, and writing a HiPS tile.
+
+    Parameters
+    ----------
+    format : `str`
+        Format of HiPS tile
+    data : `int`
```
cdeil: data is a numpy array, not an int.

hips/tiles/tile.py
```diff
@@ -0,0 +1,114 @@
...
+        Data containing HiPS tile
+    header : `format`
```
cdeil: header is an `astropy.io.fits.Header`.

adl1995:
@cdeil - I've introduced a new class `HipsTileMeta` and remove tiles from `/data` directory. I also added a utility function for getting the default path.

hips/tiles/tile.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class HipsTile:
+    """HiPS tile container.
+
+    This class provides methods for fetching, reading, and writing a HiPS tile.
+
+    Parameters
+    ----------
+    hips_tile_meta : `HipsTileMeta`
```
cdeil: Suggest to call the property just `meta`, not `hips_tile_meta`

hips/tiles/tile.py
```diff
@@ -0,0 +1,117 @@
...
+        Metadata of HiPS tile
+    format : `str`
+        Format of HiPS tile
+    data : `np.ndarray`
```
cdeil: Please always check the HTML docs locally. In this case you'll find that the link is broken and you have to put `~numpy.ndarray` instead of `np.ndarray`.

hips/tiles/tile.py
```diff
@@ -0,0 +1,117 @@
...
+        Data containing HiPS tile
+    header : `astropy.io.fits.Header`
+        Header of HiPS tile
+
+    Examples
+    --------
+        >>> import urllib.request
```
cdeil: Users should never have to import `urllib`, but just use `fetch` functions from us.

hips/tiles/tile.py
```diff
@@ -0,0 +1,117 @@
...
+        >>> from hips.tiles import HipsTile
+        >>> from hips.tiles import HipsTileMeta
+        >>> from astropy.tests.helper import remote_data
```
cdeil: Remove `remote_data` import

hips/tiles/tile.py
```diff
@@ -0,0 +1,117 @@
...
+class HipsTile:
+    """HiPS tile container.
+
+    This class provides methods for fetching, reading, and writing a HiPS tile.
+
+    Parameters
+    ----------
+    hips_tile_meta : `HipsTileMeta`
+        Metadata of HiPS tile
+    format : `str`
```
cdeil: I think you now have `format` on `meta` (which I like better). So remove from the `HipsTile` class, no?

hips/tiles/tile.py
```diff
@@ -0,0 +1,117 @@
...
+            hdulist = fits.open(raw_image)
+            data = np.array(hdulist[0].data)
+            header = hdulist[0].header
+            return cls(hips_tile_meta, data, header)
+        else:
+            data = np.array(Image.open(raw_image))
+            return cls(hips_tile_meta, data)
+
+    @classmethod
+    def read(cls, hips_tile_meta: HipsTileMeta, filename: str) -> 'HipsTile':
```
cdeil: Suggest to rename `hips_tile_meta` -> `meta` everywhere here on this class.

hips/tiles/tile.py
```diff
@@ -0,0 +1,117 @@
...
+            hdulist = fits.open(path)
+            data = np.array(hdulist[0].data)
+            header = hdulist[0].header
+            return cls(hips_tile_meta, data, header)
+        else:
+            data = np.array(Image.open(str(path)))
+            return cls(hips_tile_meta, data)
+
+    def write(self, filename: str) -> None:
+        """Write HiPS tile by a given filename (`None`).
```
cdeil: Remove `(None)` here. (no need to mention this in docstrings that don't return anything.

hips/tiles/tile.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Parameters
+        ----------
+        filename : `str`
+            Name of the file
+        """
+        path = tile_path().absolute() / filename
+        if self.hips_tile_meta.format == 'fits':
+            hdu = fits.PrimaryHDU(self.data, header=self.header)
```
cdeil: There's a `PrimaryHDU.writeto` method. No need to make an `HDUList` here.

hips/tiles/tile_meta.py
```diff
@@ -0,0 +1,29 @@
...
+
+__all__ = [
+    'HipsTileMeta',
+]
+
+
+class HipsTileMeta:
+    """HiPS tile metadata container.
+
+    This class stores HiPS tile meta attributes.
```
cdeil: Remove this line (it says the same as the summary line above).

hips/tiles/tile_meta.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Parameters
+    ----------
+    order : `int`
+        Order of HiPS tile
+    ipix : `int`
+        Pixel number of HiPS tile
+    format : `str`
+        Format of the HiPS tile
```
cdeil: Format is ambiguous. Say "file format" in the description.
Also, usually where there's a string that should be one of multiple choices, the common way to document it is this:
```
format : {'fits', 'jpeg'}
    File format
```
i.e. to list the options.

hips/utils/tile.py
```diff
@@ -0,0 +1,18 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""HEALpy wrapper functions.
```
cdeil: Copy & paste error , docstring not appropriate here.

hips/utils/tile.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+from pathlib import Path
+
+__all__ = [
+    'tile_path',
+]
+
+
+def tile_path() -> Path:
```
cdeil: This function doesn't do anything useful yet.
Do you plan to implement something here now?
If no, then I would suggest you remove this function / file for now, and only add it in a future PR if / when you need it.

cdeil:
> remove tiles from /data directory.

Note that you still have it in previous commits in this branch. I.e. there's a commit adding the files and then one removing the files. So everyone that download this code repo in the future will have to download those files.

To really get rid of the files, i.e. not introduce them into the repo, I would suggest you squash all the commits here into one. I usually use `git rebase -i master` to rebase against the latest master, but there's several ways to do this with git.

hips/utils/tile.py
adl1995: This is required for the test cases to pass, otherwise I would have to hardcode the path within `tile.py`.

adl1995:
@cdeil - I have removed `utils/tile.py` and introduced some properties in `HipsTileMeta`. Please check.

hips/tiles/tests/test_tile.py
```diff
@@ -0,0 +1,34 @@
...
+from ..tile_meta import HipsTileMeta
+
+
+class TestHipsTile:
+    def setup(self):
+        self.hips_tile_meta_fits = HipsTileMeta(order=6, ipix=30889, format='fits')
+        self.hips_tile_meta_jpg = HipsTileMeta(order=6, ipix=30889, format='jpg')
+
+    @remote_data
+    def test_fetch_read_write(self):
```
cdeil: In tests that write files, you should always write to temp directories using the pytest `tmpdir` fixture:
https://docs.pytest.org/en/latest/tmpdir.html

You can find examples here: https://github.com/astropy/astropy/search?utf8=%E2%9C%93&q=tmpdir&type=
or here: https://github.com/gammapy/gammapy/search?utf8=%E2%9C%93&q=tmpdir&type=

I would suggest we always use this pattern to make filenames:
```
def test_read_write(self, tmpdir):
    filename = str(tmpdir / 'Npix30889.jpg')
    jpeg_tile.write(filename)
    jpeg_tile2 = Tile.read(filename)
```

hips/tiles/tile.py
```diff
@@ -0,0 +1,107 @@
...
+               [0, 0, 0, ..., 0, 0, 0],
+               ...,
+               [0, 0, 0, ..., 1, 0, 0],
+               [0, 0, 0, ..., 1, 0, 1],
+               [0, 0, 0, ..., 1, 0, 1]], dtype=int16)
+        """
+
+    def __init__(self, meta: HipsTileMeta, data: np.ndarray = None, header: Header = None) -> None:
+        self.meta = meta
+        self.format = format
```
cdeil: This is a bug, you have to remove the line `self.format = format`.
The reason it isn't faliing in your tests is that `format` is a Python built-in function, i.e. exists.
That's the reason one shouldn't use variable names that match Python builtins, something a linter like pyflakes or pylint warns about.
Sometimes there's a tradeoff and you have to decide if to pick the nice name `format` or change e.g. to `fmt` to avoid this potential issue (a choice you have to make in `HipsDataMeta`). My suggestion would be to change to `file_format` in this case.

hips/tiles/tests/test_tile.py
adl1995: The data is always written to a temp directory in case of `pytest`, I think. I say this because during test start, there is a line:

`rootdir: /tmp/hips-test-kp9hkwb4,`

So, it doesn't show up in the project directory. If I were to change this to use use `tmpdir`, I would have to modify `read`, ` write` methods and pass the path along as a parameter. In the case of `classmethod` the path would then have to be stored in `self`.

hips/tiles/tile.py
```diff
@@ -0,0 +1,114 @@
...
+        if meta.file_format == 'fits':
+            hdulist = fits.open(str(path))
+            data = np.array(hdulist[0].data)
+            header = hdulist[0].header
+            return cls(meta, data, header)
+        else:
+            data = np.array(Image.open(str(path)))
+            return cls(meta, data)
+
+    def write(self, filename: str, tmpdir: py._path.local = None) -> None:
```
cdeil: The docs build is failing because of this invalid type annotation:
https://travis-ci.org/hipspy/hips/jobs/245801615#L1139

My suggestion would be that you remove this `tmpdir` argument here (and above) and instead only take in a full filename (including the dir).

adl1995:
@cdeil - I have changed the type annotation issue and passed in the full file path as the parameter. But, I'm still getting this error:

> hips/tiles/tile.py:89: error: Incompatible types in assignment (expression has type "str", variable has type "Path")

The variable `path` is only populated within the if / else block, so I'm not sure how this is an error.

cdeil:
@adl1995 - See d2ad05b . OK?

PR #26, opened 2017-06-15
Add compute_image_pixels function in hips.utils

adl1995:
I have added a function related with the computation of pixels. However, in the test case, I'm getting this error:

> urllib.error.URLError: <urlopen error An attempt was made to connect to the internet by a test that was not marked `remote_data`. The requested host was: github.com>

This is caused by the `fits.open()` function. I tried passing in a parameter called `remote_data=None` following [this](https://github.com/astropy/astropy/issues/5651) issue, but the error still persists.


hips/utils/healpix.py
```diff
@@ -54,3 +56,33 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
...
+    shape : tuple
+        Shape of the image
+    wcs : astropy.wcs.wcs.WCS
+        A WCS object containing the image header
+
+    Returns
+    -------
+    pixels : `numpy.ndarray`
+        Returns a list of pixel values
+
```
cdeil: Remove empty line here

hips/utils/healpix.py
```diff
@@ -54,3 +56,33 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
...
+    """
+
```
cdeil: Remove empty line here.

hips/utils/healpix.py
```diff
@@ -54,3 +56,33 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
...
+        The nside of the HEALPix map
+    shape : tuple
+        Shape of the image
+    wcs : astropy.wcs.wcs.WCS
+        A WCS object containing the image header
+
+    Returns
+    -------
+    pixels : `numpy.ndarray`
+        Returns a list of pixel values
```
cdeil: This description of the return value is not accurate. It's not "pixel values", but "HEALPix pixel numbers" that are computed and returned here. Pixel values would be the something like "counts" or "brightness" or whatever quantity is stored in the pixels.

hips/utils/healpix.py
```diff
@@ -54,3 +56,33 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
     boundary_coords = hp.boundaries(nside, pix, nest=nest)
     theta, phi = hp.vec2ang(np.transpose(boundary_coords))
     return theta, phi
+
+def compute_image_pixels(nside: int, shape: tuple, wcs: WCS) -> np.ndarray:
```
cdeil: Maybe rename to something like "compute_healpix_pixel_indices"?

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +25,44 @@ def test_boundaries():
     radec_precomp = [[264.375,  258.75,  264.375,  270.],
                      [-24.624318,  -30.,  -35.685335,  -30.]]
     assert_allclose([radec.ra, radec.dec], radec_precomp)
+
+def test_compute_image_pixels():
+    order = 3
+    nside = hp.order2nside(order)
+
+    hdu = fits.open('https://github.com/gammapy/gammapy-extra/blob/master/datasets/catalogs/fermi/gll_psch_v08.fit.gz?raw=true')
```
cdeil: Avoid doing web queries from tests whenever possible. They are slower and have more spurious fails than tests that just do computations and local file I/O.

Like I mentioned in the last telcon, my suggestion would be that you add a utility function to this package that generates a WCS object from scratch using a few parameters.
You can do this here or in a separate branch / PR first.

Example is here:
http://astropy.readthedocs.io/en/latest/wcs/index.html#building-a-wcs-structure-programmatically

Example how to write a function that takes a few parameters and returns a WCS object is here:
https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcs.py#L138

For this (and then later tile drawing) tests, I'd suggest to have something that doesn't cover all-sky, but maybe you could do something like this:
* AIT or CAR projection
* 1500 x 800 pixels
* Galactic coordinates with reference point at the center of the image and (GLON=10 deg, GLAT = 5 deg)

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +25,44 @@ def test_boundaries():
...
+     471, 472, 488, 489, 490, 491, 492, 493, 494, 495, 496, 497, 498, 499, 500, 501, 502, 503,
+     504, 519, 520, 521, 522, 523, 524, 525, 526, 527, 528, 529, 530, 531, 532, 533, 534, 535,
+     536, 552, 553, 554, 555, 556, 557, 558, 559, 560, 561, 562, 563, 564, 565, 566, 567, 568,
+     583, 584, 585, 586, 587, 588, 589, 590, 591, 592, 593, 594, 595, 596, 597, 598, 599, 600,
+     616, 617, 618, 619, 620, 621, 622, 623, 624, 625, 626, 627, 628, 629, 630, 631, 632, 647,
+     648, 649, 650, 651, 652, 653, 654, 655, 656, 657, 658, 659, 660, 661, 662, 663, 676, 677,
+     678, 679, 680, 681, 682, 683, 684, 685, 686, 687, 688, 689, 690, 701, 702, 703, 704, 705,
+     706, 707, 708, 709, 710, 711, 712, 713, 722, 723, 724, 725, 726, 727, 728, 729, 730, 731,
+     732, 739, 740, 741, 742, 743, 744, 745, 746, 747, 752, 753, 754, 755, 756, 757, 758, 761,
+     762, 763, 764, 765, 766, 767]
```
cdeil: Try to avoid having something like here where you have large lists or arrays of values in the test code.
It's almost always possible to find test cases where you have very small datasets that cover just as much as tests that use large lists of numbers or lots of computations.

If the result value is still long (say more than 10 numbers), then you could just assert on the first and last value and length of the array, instead of copy & pasting 100s of numbers into the test code file.

adl1995:
@cdeil I have added a class called `WCSGeometry` and renamed the function to `compute_healpix_pixel_indices`. For now, the class just contains two properties, `wcs` and `shape`, but I will add a class method to create a WCS object within this class.

adl1995:
@cdeil - I added a `@classmethod` for creating a WCS object, however, I'm getting this error in `wcs.all_pix2world` function:

> TypeError: WCS projection has 4 dimensions, so expected 2 (an Nx4 array and the origin argument) or 5 arguments (the position in each dimension, and the origin argument). Instead, 3 arguments were given.

I think this has something to do with how I'm constructing my `SkyCoord` object. For now, I am passing it these arguments:

`skycoord = SkyCoord(10, 20, unit="deg")`

hips/utils/wcs.py
```diff
@@ -33,17 +34,53 @@ def wcs(self) -> None:
...
+        cdelt : `float`
+            Coordinate increment at reference point
+        crpix : `float`
+            Pixel coordinate of reference point
         """
-        # TODO
-        pass
+
+        naxis = 2
+        if shape is not None:
```
cdeil: At least for now, we only process 2-dim images.
So here you can just remove the part computing naxis from shape, and instead just do this:
```
w = WCS(naxis=2)
```
and then the WCS transforms should work as you know them from other examples / your notebooks

hips/utils/wcs.py
```diff
@@ -0,0 +1,86 @@
...
+
+class WCSGeometry:
+    """Container for WCS object and image shape.
+
+    Parameters
+    ----------
+    wcs : `~astropy.wcs.WCS`
+        WCS projection object
+    shape : tuple
+        Shape of the image
```
cdeil: Suggest to add an `Examples` section here, showing the example you have in the tests, i.e. how to call `wcs_geo = WCSGeometry.create`, and then print `wcs_geo.wcs` and `wcs_geo.shape` to show it's content.

hips/utils/wcs.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    def __init__(self, wcs: WCS, shape: tuple) -> None:
+        self._wcs = wcs
+        self._shape = shape
+
+    @property
+    def shape(self) -> None:
```
cdeil: Suggest to remote the `property` for shape and wcs, and to instead store `self.wcs=wcs` and `self.shape` directly in `__init__`.

hips/utils/wcs.py
```diff
@@ -0,0 +1,86 @@
...
+        return self._shape
+
+    @property
+    def wcs(self) -> None:
+        """WCS object containing FITS image header (`None`)."""
+        return self._wcs
+
+    @classmethod
+    def create(cls, skydir: SkyCoord, shape: tuple, coordsys: str='CEL', projection: str='AIT', cdelt: float=1.0, crpix: float=1.) -> 'WCSGeometry':
+        """Read from HiPS description file (`WCSGeometry`).
```
cdeil: This summary line in the docstring was copied over from some other class. Please update it to say something appropriate for this class.

hips/utils/wcs.py
```diff
@@ -0,0 +1,86 @@
...
+        """Shape of the image (`None`)."""
+        return self._shape
+
+    @property
+    def wcs(self) -> None:
+        """WCS object containing FITS image header (`None`)."""
+        return self._wcs
+
+    @classmethod
+    def create(cls, skydir: SkyCoord, shape: tuple, coordsys: str='CEL', projection: str='AIT', cdelt: float=1.0, crpix: float=1.) -> 'WCSGeometry':
```
cdeil: Wrap this line at ~ 80 char.

hips/utils/wcs.py
```diff
@@ -0,0 +1,86 @@
...
+            Sky coordinate of the WCS reference point
+        shape : `tuple`
+            Shape of the image
+        coordsys : `str`
+            Coordinate system
+        projection : `str`
+            Projection of the WCS object
+        cdelt : `float`
+            Coordinate increment at reference point
+        crpix : `float`
```
cdeil: I see below in this function you also support the case of having a tuple for `crpix`.
The single float case is a bit weird, usually one wants to set the reference point in x and y, i.e. the general case.
Suggest to remove support for the single float `crpix` (remove the try-except below) and always require the user to give a tuple of floats here.

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +24,19 @@ def test_boundaries():
...
+def test_compute_healpix_pixel_indices():
+    order = 3
+    nside = hp.order2nside(order)
+
+    skycoord = SkyCoord(10, 20, unit="deg")
+    wcs_geometry = WCSGeometry.create(skycoord, (10, 20), 'CEL', 'AIT', 1.0, 1)
+
+    """
+    These pixel values were obtained for the all-sky image located at:
+    https://github.com/gammapy/gammapy-extra/blob/master/datasets/catalogs/fermi/gll_psch_v08.fit.gz?raw=true
```
cdeil: I think this comment no longer applies -> remove.

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +24,19 @@ def test_boundaries():
...
+    """
+    pixels_precomp = [0, 767]
+
+    pixels = compute_healpix_pixel_indices(wcs_geometry, nside)
+    assert_allclose([pixels[0], pixels[-1]], pixels_precomp)
```
cdeil: Can you add an assert on `len(pixels)`? How many are there now for this example you have.
You could also change the example so that you have ~ 5 to 10 pixels, and then just do the `assert_allclose` on all of those.
I was only objecting to the ~ 700 pixels long list you had before, because it made the test hard to read.

hips/utils/healpix.py
```diff
@@ -50,7 +53,32 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
         [( 264.375, -24.62431835), ( 258.75 , -30.        ),
          ( 264.375, -35.68533471), ( 270.   , -30.        )]>
     """
-
     boundary_coords = hp.boundaries(nside, pix, nest=nest)
     theta, phi = hp.vec2ang(np.transpose(boundary_coords))
     return theta, phi
+
+def compute_healpix_pixel_indices(wcs_geometry: WCSGeometry, nside: int) -> np.ndarray:
+    """Returns an array containing pixels corresponding to an image.
```
cdeil: > Returns an array containing pixels corresponding to an image.

I don't think this is quite accurate. Usually the return list of HEALPix pixels will contain pixels completely outside the image, no? (because you're doing a disk query and images don't correspond to disk regions. Suggest to change the description a bit to be more precise on that this function does.

adl1995:
@cdeil - I just made a commit which adds 2 examples, updates the test case, `crpix` parameter for `WCSGeometry`, and some documentation changes.

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +24,17 @@ def test_boundaries():
...
+    order = 3
+    nside = hp.order2nside(order)
+
+    skycoord = SkyCoord(10, 20, unit="deg")
+    wcs_geometry = WCSGeometry.create(skydir=skycoord, shape=(10, 20), coordsys='CEL', projection='AIT', cdelt=1.0, crpix=(1., 1.))
+
+    pixels_precomp = [84, 111, 112, 113]
+    len_precomp = 39
+    pixels = compute_healpix_pixel_indices(wcs_geometry, nside)
+    print(pixels)
```
cdeil: Remove print here. Do this always before pushing to Github. There should never be print statements in library code like hips or in the tests.

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +24,17 @@ def test_boundaries():
     radec_precomp = [[264.375,  258.75,  264.375,  270.],
                      [-24.624318,  -30.,  -35.685335,  -30.]]
     assert_allclose([radec.ra, radec.dec], radec_precomp)
+
+def test_compute_healpix_pixel_indices():
+    order = 3
+    nside = hp.order2nside(order)
+
+    skycoord = SkyCoord(10, 20, unit="deg")
+    wcs_geometry = WCSGeometry.create(skydir=skycoord, shape=(10, 20), coordsys='CEL', projection='AIT', cdelt=1.0, crpix=(1., 1.))
```
cdeil: Wrap this line. It's too long, I have to scroll back and forth on Github to read it.

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +24,17 @@ def test_boundaries():
...
+
+    pixels_precomp = [84, 111, 112, 113]
+    len_precomp = 39
```
cdeil: For cases like here where you just define a local variable `len_precomp` and use it once on a line that isn't very long, I'd suggest o just use the value in the assert, i.e. not introduce the local variable.

Personally I do introduce local variables to make code more readable when lines get too long. But in cases like here where the assert statement is very simple and the line is short, no need for the local variable.

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +24,17 @@ def test_boundaries():
...
+    pixels = compute_healpix_pixel_indices(wcs_geometry, nside)
+    print(pixels)
+    assert_allclose(len(pixels), len_precomp)
```
cdeil: Just use this here:
```
assert len(pixels) == 39
```

hips/utils/tests/test_healpix.py
```diff
@@ -22,3 +24,17 @@ def test_boundaries():
     radec_precomp = [[264.375,  258.75,  264.375,  270.],
                      [-24.624318,  -30.,  -35.685335,  -30.]]
     assert_allclose([radec.ra, radec.dec], radec_precomp)
+
+def test_compute_healpix_pixel_indices():
+    order = 3
+    nside = hp.order2nside(order)
```
cdeil: I would suggest to move this down to where you use it, i.e. above the `compute_healpix_pixel_indices` line.
Also: you don't need the `order` local variable, i.e. just spare a line and write:
```
nside = hp.order2nside(order=3)
```

hips/utils/wcs.py
```diff
@@ -0,0 +1,90 @@
...
+        Shape of the image
+
+    Examples
+    --------
+    ::
+
+        >>> from hips.utils import WCSGeometry
+        >>> from astropy.coordinates import SkyCoord
+        >>> skycoord = SkyCoord(10, 20, unit="deg")
+        >>> wcs_geometry = WCSGeometry.create(skydir=skycoord, shape=(10, 20), coordsys='CEL', projection='AIT', cdelt=1.0, crpix=1.)
```
cdeil: Line is too long to read -> wrap it.

hips/utils/healpix.py
```diff
@@ -50,7 +53,49 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
...
+        >>> order = 3
+        >>> nside = hp.order2nside(order)
+        >>> skycoord = SkyCoord(10, 20, unit="deg")
+        >>> wcs_geometry = WCSGeometry.create(skydir=skycoord, shape=(10, 20), coordsys='CEL', projection='AIT', cdelt=1.0, crpix=(1., 1.))
+        >>> compute_healpix_pixel_indices(wcs_geometry, nside)
+        [ 84 111 112 113 142 143 144 145 146 174 175 176 177 178 206 207 208 209
+        210 238 239 240 241 270 271 272 273 274 302 303 304 305 334 335 336 337
+        367 368 399]
+    """
+    y_center, x_center = wcs_geometry.shape[0] // 2, wcs_geometry.shape[1] // 2
```
cdeil: Actually, I think here it should be `/` (float division) and not `//` (integer devision, which I think I suggested before).

hips/utils/healpix.py
```diff
@@ -50,7 +53,49 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
...
+    lon_center, lat_center = wcs_geometry.wcs.all_pix2world(x_center, y_center, 1)
```
cdeil: The third `origin` parameter here should be `0`, not `1`.
The logic is to use `0` when pixel coordinates are 0-based, like with numpy arrays, and to always use 0-based pixel coordinates in our code.
See https://photutils.readthedocs.io/en/stable/photutils/overview.html#coordinate-conventions

hips/utils/healpix.py
```diff
@@ -50,7 +53,49 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
...
+    vec = hp.ang2vec(lon_center, lat_center, lonlat=True)
```
cdeil: I'd suggest to move this line `vec` down, just above the `hp.query_disc` call.
I.e. have the separation computations, which are independent of healpy come first.
We might split this up into two functions, but for now it's OK to keep in one function.

hips/utils/wcs.py
```diff
@@ -0,0 +1,90 @@
...
+        PC2_1 PC2_2  : 0.0  1.0
+        CDELT : -1.0  1.0
+        NAXIS : 0  0
+        >>> wcs_geometry.shape
+        (10, 20)
+    """
+
+    def __init__(self, wcs: WCS, shape: tuple) -> None:
+        self.wcs = wcs
+        self.shape = shape
```
cdeil: @adl1995 - how about introducing a `namedtuple` for shape and convert the given shape to a `Shape` here?
Then everywhere else in the code we could use `shape.nx` and `shape.ny`, which is more readable than using `[0]` and `[1]` where one always has to think which is which, no?

```
In [1]: shape = (10, 20)

In [2]: from collections import namedtuple

In [3]: namedtuple('Shape', ['ny', 'nx'])
Out[3]: __main__.Shape

In [4]: Shape = namedtuple('Shape', ['ny', 'nx'])

In [5]: Shape((10, 20))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-44cf43da6338> in <module>()
----> 1 Shape((10, 20))

TypeError: __new__() missing 1 required positional argument: 'nx'

In [6]: Shape(10, 20)
Out[6]: Shape(ny=10, nx=20)

In [7]: Shape(10, 20).nx
Out[7]: 20

In [8]: Shape(10, 20).ny
Out[8]: 10

In [9]: Shape(*shape)
Out[9]: Shape(ny=10, nx=20)
```

adl1995:
@cdeil - I've made the necessary changes. Please check.

hips/utils/healpix.py
```diff
@@ -50,7 +55,51 @@ def boundaries(nside: int, pix: int, nest: bool=True) -> tuple:
...
+coordsys='CEL', projection='AIT', \
+cdelt=1.0, crpix=(1., 1.))
+        >>> compute_healpix_pixel_indices(wcs_geometry, nside)
+        [ 84 111 112 113 142 143 144 145 146 174 175 176 177 178 206 207 208 209
+        210 238 239 240 241 270 271 272 273 274 302 303 304 305 334 335 336 337
+        367 368 399]
+    """
+    y_center, x_center = wcs_geometry.Shape.ny / 2, wcs_geometry.Shape.nx / 2
+    lon_center, lat_center = wcs_geometry.wcs.all_pix2world(x_center, y_center, 1)
+    separations = angular_separation(x_center, y_center, lon_center, lat_center)
```
cdeil: This function still wasn't working. You forgot to compute the grid of sky coordinates of the pixels (starting with `np.indices` and then calling `pix2world` like we discussed in the last telcon). I've fixed this and a few other things, will push my commit soon.

cdeil:
@adl1995 - See 17925ac . I'll merge if travis-ci passes.

Some things I did:
- (Hopefully) fix the `compute_healpix_pixel_indices` function
- Add methods and some tests for `WCSGeometry`
- Some misc code and docstring cleanup

PR #27, opened 2017-06-19
Add class HipsSurveyList for parsing list of hips surveys

adl1995:
I added a class for creating a `astropy.table.Table` from a text file. It reads a list of `OrderedDict` and creates a table from it.


hips/tiles/surveys.py
```diff
@@ -0,0 +1,62 @@
...
+    def parse(cls, text: str) -> 'HipsSurveyList':
+        """Parse HiPS list text (`HipsSurveyList`).
+
+        Parameters
+        ----------
+        text : str
+            HiPS list text
+        """
+        surveys = list()
+        print(len(text.split('\n\n')))
```
cdeil: remove print statement.

hips/tiles/tests/test_surveys.py
```diff
@@ -0,0 +1,14 @@
...
+
+class TestHipsSurveyList:
+    @classmethod
+    def setup_class(cls):
+        filename = get_pkg_data_filename('data/surveys.txt')
+        cls.hips_survey_list = HipsSurveyList.read(filename)
+
+    def test_surveys(self):
+        len_precomp = 335
+        assert len(self.hips_survey_list.surveys) == len_precomp
```
cdeil: I'd suggest to change the test file to just have a few entries (e.g. 3).
Also, please add an assert or two on some data value in the table, which shows that it was parsed correctly.

hips/tiles/surveys.py
```diff
@@ -0,0 +1,62 @@
...
+        for raw_survey in text.split('\n\n'):
+            survey = OrderedDict()
+            for line in raw_survey.split('\n'):
```
cdeil: Isn't this duplicating the code we already have in https://hips.readthedocs.io/en/latest/api/hips.HipsDescription.html#hips.HipsDescription.parse
Can you call that instead to do the parsing of each survey?

adl1995:
@cdeil - I've made some changes. But, the `parse` method still duplicates the code from `HipsDescription`. Is there any way I can call the `parse` `@classmethod` from within `HipsTile`?

cdeil:
My understanding is that the `HipsSurveyList` created here is a list of
https://hips.readthedocs.io/en/latest/api/hips.HipsDescription.html
and the text file format for `HipsSurveyList` is a bunch of text blocks matching the text file format for a single `HipsDescription` concatenated, separated by an empty line.

Is this correct or not?

If yes, you should get the text blocks by splitting on empty lines as you do, and then call `HipsDescription.parse` on each of these text blocks. OK?

adl1995:
@cdeil - I have renamed the class `HipsDescription` and `HipsSurveyList` to `HipsSurveyList` and `HipsSurveyPropertyList`. I've also updated the `parse` method to use `HipsSurveyList.parse`. Please check.

hips/tiles/tests/test_surveys.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @remote_data
+    def test_fetch(self):
+        self.hips_survey_list.fetch()
+
+    def test_surveys(self):
+        len_precomp = 4
+
+        assert len(self.hips_survey_list.surveys) == len_precomp
+        assert self.hips_survey_list.surveys[0]['surveys']['creator_did'] == 'ivo://CDS/C/MUSE-M42'
```
cdeil: You could introduce a temp variable here to keep the assert statements nice and short:
```
survey = self.hips_survey_list.surveys[0]['surveys']
```
Also: accessing [0]['surveys'] is very weird, suggest to not put the data in a Table, but just a list of `HipsSurveyProperties` objects.

And please also add a similar assert for some entry other than the first one. E.g. for the last survey you have in that example file.

hips/tiles/tests/test_surveys.py
```diff
@@ -0,0 +1,22 @@
...
+    def setup_class(cls):
+        filename = get_pkg_data_filename('data/surveys.txt')
+        cls.hips_survey_list = HipsSurveyPropertyList.read(filename)
+
+    @remote_data
+    def test_fetch(self):
+        self.hips_survey_list.fetch()
+
+    def test_surveys(self):
+        len_precomp = 4
```
cdeil: Please don't use the "precomp" name, it's not very common. Common generic names are "expected" or "desired". In this case though, I think don't need the local variable (just used once, on a line that's already short, just copy the value here).

hips/tiles/surveys.py
```diff
@@ -58,6 +58,10 @@ def parse(cls, text: str) -> 'HipsSurveyPropertiesList':
         surveys = list()
         for raw_survey in text.split('\n\n'):
             surveys.append(HipsSurveyProperties.parse(raw_survey).properties)
+        return cls(surveys)
+
+    @property
+    def table(self) -> Table:
```
cdeil: Every public property needs at least a one-line docstring.
And there should be a test covering everything, i.e. in this case for the table.

Also -- I think this way of creating a Table isn't useful. Can you do `table.info()` to show what it contains? Is it a single column? You can't do any selection like that, no?

Probably what you want is to have a table with one row per survey, and the properties as columns.
Can you try `table = Table(rows=rows)` and pass as `rows` a Python list of `OrderedDict` objects?
That should populate the table in a useful way, i.e. with multiple columns and rows as I mentioned above.

If this doesn't work, you could also remove the table property here and then add that in a future PR.

hips/tiles/surveys.py
```diff
@@ -14,11 +14,11 @@ class HipsSurveyPropertiesList:
 
     Parameters
     ----------
-    surveys : `~astropy.table.Table`
+    surveys : list
         HiPS surveys
     """
 
-    def __init__(self, surveys: Table) -> None:
+    def __init__(self, surveys: list) -> None:
```
cdeil: I think  for type annotations to be useful, you'd have to use `from typing import List` and declare the type as `List[HipsSurveyProperties]` here. I've never used that, but maybe you can try?

adl1995:
@cdeil - I have modified the type to `List[HipsSurveyProperties]` in the `__init__` method, but this gives me an error when running `mypy`:

> hips/tiles/surveys.py:61: error: Argument 1 to "HipsSurveyPropertiesList" has incompatible type List[OrderedDict[Any, Any]]; expected List[HipsSurveyProperties]

So, I haven't pushed my changes yet. I don't have much experience with `mypy`, but I tried changing all the type annotations from `HipsSurveyPropertiesList` to `List[HipsSurveyProperties]`. This removes all the errors expect for this one:

> hips/tiles/surveys.py:61: error: Incompatible return value type (got "HipsSurveyPropertiesList", expected List[HipsSurveyProperties])

> hips/tiles/surveys.py:61: error: Argument 1 to "HipsSurveyPropertiesList" has incompatible type List[OrderedDict[Any, Any]]; expected List[HipsSurveyProperties]

hips/tiles/surveys.py
```diff
@@ -0,0 +1,67 @@
...
+    @classmethod
+    def read(cls, filename: str) -> 'HipsSurveyPropertiesList':
+        """Read HiPS list from file (`HipsSurveyPropertiesList`).
+
+        Parameters
+        ----------
+        filename : str
+            HiPS list filename
+        """
+        with open(filename, 'r', encoding='utf-8', errors='ignore') as fh:
```
cdeil: 'r' is the default mode for open, no need to specify it.

hips/tiles/surveys.py
```diff
@@ -0,0 +1,67 @@
...
+    @classmethod
+    def fetch(cls, url: str='http://alasky.unistra.fr/MocServer/query?hips_service_url=*&dataproduct_type=!catalog&dataproduct_type=!cube&get=record') -> 'HipsSurveyPropertiesList':
+        """Fetch HiPS list text from remote location (`HipsSurveyPropertiesList`).
+
+        Parameters
+        ----------
+        url : str
+            HiPS list URL
+        """
+        text = urllib.request.urlopen(url).read()
```
cdeil: I would suggest to call `decode` instead of calling `str(text), i.e. change to this to be more clear what is happening:
```
response = urllib.request.urlopen(url).read()
text = response.decode('utf-8')
return cls.parse(text)
```

hips/tiles/surveys.py
```diff
@@ -0,0 +1,67 @@
...
+        """Parse HiPS list text (`HipsSurveyPropertiesList`).
+
+        Parameters
+        ----------
+        text : str
+            HiPS list text
+        """
+        surveys = list()
+        for raw_survey in text.split('\n\n'):
+            surveys.append(HipsSurveyProperties.parse(raw_survey).properties)
```
cdeil: I think the mypy error you mentioned might be due to this line.
You're calling `.properties` here, i.e. storing `OrderedDict` objects in the list.
What do you want to store as data member on `HipsSurveyPropertiesList`.
I would suggest you store `list` of `HipsSurveyProperties`, and not `list` of `OrderedDict`, i.e. don't remove the `.properties` here.

adl1995:
@cdeil - I tried `response.decode('utf-8')`, but this gives me an error:

> UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb2 in position 83958: invalid start byte

Also, I think you meant to remove the `.properties` from `HipsSurveyProperties`, this removes the `mypy` error, but I'm unable to use Astropy `Table` with this.

hips/tiles/description.py
```diff
@@ -1,26 +1,38 @@
...
     ----------
     properties : `~collections.OrderedDict`
-        HiPS description properties
+        HiPS survey properties
+
+    Examples
+    --------
+    ::
+
+        >>> from astropy.utils.data import get_pkg_data_filename
```
cdeil: You're still using `get_pkg_data_filename` from a docs example here. As discussed previously, this won't work. Please change to a working example where you instead fetch a file from Github or CDS.

hips/tiles/description.py
```diff
@@ -2,25 +2,37 @@
...
     properties : `~collections.OrderedDict`
-        HiPS description properties
+        HiPS survey properties
+
+    Examples
+    --------
+    ::
+
+        >>> import urllib.request
+        >>> response = urllib.request.urlopen('https://raw.githubusercontent.com/hipspy/\
```
cdeil: This docs example is failing at the moment:
https://gist.github.com/cdeil/422a87d598b1e60e7e5c6bff25d1b716

Please add a `fetch` method to this `HipsSurveyProperties` class and call it from the docs example.

adl1995:
@cdeil - I added the `fetch` classmethod, but also re-based it against `master`, so there are also other files changes showing up.

cdeil:
@adl1995 - I think for this PR you also rebased against some old master branch, so now there's a lot of stuff in the diff unrelated to this PR.

Please rebase against latest master and squash all commits into a single one with a good commit message.

cdeil:
@adl1995 - I've added a commit in c12acd9ab7797cd7fd7e08a80f0ee802715e95f3 that fixes HipsSurveyPropertiesList.fetch (conversion to text and parsing was broken) and adds a test for that,  adds `from .surveys import *` to `__init__` so that the class shows up in the docs, as well as a bunch of cleanup. I did rename the data members on the `HipsSurveyProperties` and `HipsSurveyPropertiesList` classes to just "data", because I found the previous names confusing because those were often the names of the local variable (i.e. `survey.survey` or `properties.properties` is confusing).

@adl1995 - Will merge soon when travis-ci passes, unless you have any comments.

PR #30, opened 2017-06-23
High-level function: make_sky_image

adl1995:
Added a function `make_sky_image` and passed it by adding the `@pytest.mark.xfail` decorator. The high level functionality for this is yet to be discussed.


cdeil:
@adl1995 - Please make a first draft of this function, i.e. actually write inputs, code and return something, and write the docstring and test.

There's some info here:
https://github.com/hipspy/hips/blob/master/notes/API-proposal.md#make-a-wcs-image
and we also discussed it a few weeks ago here:
https://docs.google.com/document/d/1D3bEJuSH4d99u9dH_hGur3jbbPNapRI2hduYHCH9NKk/edit#heading=h.paw124lueu5w
(the ` make_sky_image(geometry, hips_survey)`).

By now, we've introduced the `WCSGeometry` class, that should be the "geometry" input mentioned.

If you want, you could also start to add classes for the missing functionality here. I mean just put the classes with a docstring and the methods you need and put "raise NotImplementedError" in the methods.

So what we do here in this PR is to actually write down the API we want as actual Python functions / classes, they just aren't implemented yet. Then we'll merge, and you'll implement the missing pieces in small PRs one at a time next week and I think by the end of next week, we'll have something we can call v0.1 and release. :-)

cdeil:
I see you've renamed `hips/draw/simple.py → hips/draw/tile.py` here.
I'd suggest you don't introduce `hips/draw/tile.py` and keep as-is.
All of our drawing / functionality will be tile-related.
The idea with `simple.py` was that this is where you implement https://hips.readthedocs.io/en/latest/drawing_algo.html#naive-algorithm and then https://hips.readthedocs.io/en/latest/drawing_algo.html#precise-algorithm will go into some other file, like `precise.py`.

We might also need a `high_level.py` or something like that with a uniform interface into the different drawing methods. I'd be OK if you add `high_level.py` now if you like, and only have the simple tile drawing stuff in `simple.py`.

hips/draw/simple.py
```diff
@@ -1,5 +1,24 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """HiPS tile drawing -- simple method."""
 
+import healpy as hp
```
cdeil: We said we would use `hips/utils/healpix.py` as wrapper and only import `healpy` there, to make it possible to eventually completely get rid of `healpy`.
I see two options: either expose `hips_nside` as a property (computed from `hips_order`) on the `HipsSurveyProperties` object, or change `compute_healpix_pixel_indices` to take `order` instead of `nside` as parameter (since `order` seems to be the more common quantity available in HipS).

adl1995:
@cdeil - I just made a commit adding the missing code. I made an assert on `[0, 0]`. Although it is passing, I don't think the result is okay. I tried displaying the resulting image through `matplotlib` and all the image is black except for a small portion at the top. This leads me to think that the tiles were outside the sky image. Could you please take a look?

adl1995:
Also, the resulting image of `test_draw_sky_image` was all zeros, so I marked it with `pytest.mark.xfail`.

hips/draw/simple.py
```diff
@@ -1,5 +1,158 @@
...
+        An object of WCSGeometry
+    tiles : List[HipsTile]
+        A list of HipsTile
+
+    Returns
+    -------
+    np.ndarray
+        Returns a numpy array containing all HiPS tiles projected onto it
+    """
+    # TODO: copy over (and clean up a little) the drawing code you already have in notebooks.
```
cdeil: Remove this TODO comment?

hips/draw/simple.py
```diff
@@ -1,5 +1,158 @@
...
+    # Suggestion: don't debug the code now, just put it here.
+    # First implement simple tile caching so that you can draw without needing to make web queries.
+
+    # Probably this function should just loop over tiles and call a `_draw_sky_image_one_tile`
+    # helper function, and sum up the results.
+    all_sky = np.zeros(geometry.shape)
+    for tile in tiles:
+        nside = hp.order2nside(tile.meta.order)
```
cdeil: Here you're already starting to do computations that are specific to this one tile.
Those should be moved to the `_draw_sky_image_one_tile` function called below, no?

adl1995:
@cdeil - I moved the `_draw_sky_image_one_tile` function to `DrawSkyImageOneTile` class. Also, I verified the output through `matplotlib` and the result looks okay.

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """HiPS tile drawing -- simple method."""
+import os
+from pathlib import Path
+
+import healpy as hp
```
cdeil: Hmmm ... we said that we would try to `import healpy` only from `hips.utils.healpix`.
I think here you're only accessing it from a method that would better fit on `TileMeta`, or maybe it can go to `hips.utils.healpix` as an extra wrapper function, or even replacing the existing function that computes corners?

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+    'make_sky_image',
 ]
+
+
+def draw_sky_image(geometry: WCSGeometry, tiles: List[HipsTile]) -> np.ndarray:
+    """Draw sky image using the simple and quick method.
+
+    Parameters
+    ----------
+    geometry : 'WCSGeometry'
```
cdeil: Can you please check the output HTML docs?
Does this generate a link to `WCSGeometry`?
I think you have to change to this, no?
```
geometry : `~hips.utils.WCSGeometry`
```

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+    all_sky = np.zeros(geometry.shape)
+    for tile in tiles:
+        draw_tile = DrawSkyImageOneTile(geometry, tile, all_sky.shape)
+        draw_tile.compute_corners()
+        draw_tile.apply_projection()
+        all_sky += draw_tile.warp_image()
+    return all_sky
+
+
+class DrawSkyImageOneTile:
```
cdeil: Maybe rename to `SimpleTilePainter`?

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+    """A private function for drawing a tile over a sky image.
```
cdeil: "a private function" -> this is not a function, and not private. Put a better one-line summary.

Also, please link to https://hips.readthedocs.io/en/latest/drawing_algo.html#naive-algorithm and say that this class implements the method described there.

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+
+    Parameters
+    ----------
+    geometry : 'WCSGeometry'
+        An object of WCSGeometry
+    tile : HipsTile
+       An object of HipsTile
+    shape ; tuple
```
cdeil: More docstring formatting issues: `;` instead of `:`.
Please always run tests and build and check the docs locally before pushing to Github.

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+        Shape of the all-sky image
+    """
+
+    def __init__(self, geometry: WCSGeometry, tile: HipsTile, shape: tuple) -> None:
+        self.geometry = geometry
+        self.tile = tile
+        self.shape = shape
+        self.wcs = WCS(tile.header)
```
cdeil: Remove this `self.wcs`? It's currently not working, not documented and not used, no?

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+        nside = hp.order2nside(self.tile.meta.order)
+        theta, phi = boundaries(nside, self.tile.meta.ipix)
+        radec = SkyCoord(ra=phi, dec=np.pi / 2 - theta, unit='radian', frame='icrs')
+        self.corners = []
+        for i in range(len(radec.ra.deg)):
+            self.corners.append([radec.ra.deg[i], radec.dec.deg[i]])
+
+    def apply_projection(self) -> None:
+        src = self.geometry.wcs.wcs_world2pix(self.corners, 0)
+        dst = np.array([[511, 0], [511, 511], [0, 511], [0, 0]])
```
cdeil: The number 511 shouldn't be hard-coded, but should be computed from the tile `shape`, no?
My understanding is that it is usually, but not always 512 x 512 pixels.
Maybe this `dst` array would make sense as a property on `HipsTileMeta`?

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+    def compute_corners(self) -> None:
+        nside = hp.order2nside(self.tile.meta.order)
+        theta, phi = boundaries(nside, self.tile.meta.ipix)
+        radec = SkyCoord(ra=phi, dec=np.pi / 2 - theta, unit='radian', frame='icrs')
+        self.corners = []
+        for i in range(len(radec.ra.deg)):
+            self.corners.append([radec.ra.deg[i], radec.dec.deg[i]])
+
+    def apply_projection(self) -> None:
+        src = self.geometry.wcs.wcs_world2pix(self.corners, 0)
```
cdeil: I would suggest to just store `self.corners` as a SkyCoord of length 4, and then to call http://astropy.readthedocs.io/en/latest/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.to_pixel

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+
+    def compute_corners(self) -> None:
+        nside = hp.order2nside(self.tile.meta.order)
+        theta, phi = boundaries(nside, self.tile.meta.ipix)
+        radec = SkyCoord(ra=phi, dec=np.pi / 2 - theta, unit='radian', frame='icrs')
+        self.corners = []
+        for i in range(len(radec.ra.deg)):
+            self.corners.append([radec.ra.deg[i], radec.dec.deg[i]])
+
+    def apply_projection(self) -> None:
```
cdeil: This method is badly named. It **computes** the transformation, it doesn't **apply** it.
Maybe call it `compute_projection`?

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+        HiPS survey properties
+
+    Returns
+    -------
+    data : `~numpy.ndarray`
+        Output image pixels
+
+    Examples
+    --------
+    >>> from astropy.io import fits
```
cdeil: Is this example working now?
If no, can you try to put a working example?
Or is there still some functionality missing?

hips/draw/simple.py
```diff
@@ -1,5 +1,156 @@
...
+    ... )
+    >>> url = 'https://raw.githubusercontent.com/hipspy/hips-extra/master/datasets/samples/DSS2Red/properties'
+    >>> hips_survey = HipsSurveyProperties.fetch(url)
+    >>> data = make_sky_image(geometry, hips_survey)
+    >>> hdu = fits.PrimaryHDU(data=data, header=geometry.header)
+    >>> hdu.writeto('my_image.fits')
+    """
+    healpix_pixel_indices = compute_healpix_pixel_indices(geometry, hips_survey.hips_order)
+    path = Path(os.environ['HIPS_EXTRA'])
+    tiles_path = path / 'datasets' / 'samples' / 'DSS2Red' / 'Norder3' / 'Dir0'
```
cdeil: Clearly hard-coding "DSS2Red" isn't a good solution.
Can't you change the API of this function so that the user can select what they want?
Alternatively, if you want to do that in a follow-up PR, at least leave a "TODO" comment here.

hips/draw/simple.py
adl1995: I just tried this, but the two results vary. Here is the result of:

```
radec = SkyCoord(ra=phi, dec=np.pi / 2 - theta, unit='radian', frame='icrs')
self.corners = []
for i in range(len(self.corners.ra.deg)):
    self.corners.append([radec.ra.deg[i], radec.dec.deg[i]])
```

> [[0.0, 9.5940682268604593], [354.375, 4.78019184719916], [0.0, 0.0], [5.625, 4.78019184719916]]

However using [SkyCoord.to_pixel](http://astropy.readthedocs.io/en/latest/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.to_pixel) i.e. `self.corners.to_pixel(self.geometry.wcs), 0)` where `self.corners` only contains the `SkyCoord` object gives this output:

> [[-5706.15819396 -4779.65912225 -4203.26122367 -5237.6912678 ]
>  [-4853.94661884 -4966.10931322 -5593.68414642 -5457.21476509]]

cdeil: I'm not sure why the results are different. Do you know which is correct?

Can you make a self-contained Jupyter notebook with this WCS and tile where the tile corner sky and pixel coordinates are printed and plotted. I know this is quite some work, but I think it's needed to get correct results here and for me to be able to review the tests in this PR, i.e. you should show clearly via the tests and in addition the Jupyter notebook that this is giving correct results.

hips/draw/simple.py
tboch: you are right, tile size can be different from 512x512 pixels

hips/draw/simple.py
```diff
@@ -1,5 +1,152 @@
...
+    for tile in tiles:
+        draw_tile = SimpleTilePainter(geometry, tile, all_sky.shape)
+        draw_tile.compute_corners()
+        draw_tile.compute_projection()
+        all_sky += draw_tile.warp_image()
+    return all_sky
+
+
+class SimpleTilePainter:
+    """A class which iteratively draws a tile following the naive algorithm steps mentioned `here <https://hips.readthedocs.io/en/latest/drawing_algo.html#naive-algorithm>`_.
```
cdeil: Please always omit fill words like "a class". In this case I would suggest the following docstring:
```
"""
Paint a single tile using a simple projective transformation method.

The algorithm implemented is described here: :ref:`drawing_algo`.
```
Note how I linked to the high-level docs with `:ref:` pointing to a label there. Within our docs this is always how cross-links should be made.

hips/draw/simple.py
```diff
@@ -1,5 +1,152 @@
...
+
+    Parameters
+    ----------
+    geometry : `~hips.utils.WCSGeometry`
+        An object of WCSGeometry
+    tile : `HipsTile`
+       An object of HipsTile
+    shape : tuple
```
cdeil: The `shape` is an attribute of `geometry`. No need to pass `shape` separately, no?

hips/draw/simple.py
```diff
@@ -1,5 +1,152 @@
...
+    """
+
+    def __init__(self, geometry: WCSGeometry, tile: HipsTile, shape: tuple) -> None:
+        self.geometry = geometry
+        self.tile = tile
+        self.shape = shape
+        self.corners = None
+        self.pt = None
+
+    def compute_corners(self) -> None:
```
cdeil: Suggest to make this a property:
```
@property
def corner_pixel_coordinates(self)
    ... mostly the code you have now ...
    return the result
```

hips/draw/simple.py
```diff
@@ -1,5 +1,152 @@
...
+        theta, phi = boundaries(self.tile.meta.nside, self.tile.meta.ipix)
+        radec = SkyCoord(ra=phi, dec=np.pi / 2 - theta, unit='radian', frame='icrs')
```
cdeil: Here you hardcode `frame='icrs'`. But tiles can be in Galactic frame.
So here you should instead add the frame as an attribute to `HipsTileMeta` and then access it from here.

hips/draw/tests/test_simple.py
```diff
@@ -1 +1,52 @@
...
+from astropy.tests.helper import remote_data
+from numpy.testing import assert_allclose
+
+from ..simple import make_sky_image, draw_sky_image
+from ...tiles import HipsSurveyProperties, HipsTileMeta, HipsTile
+from ...utils.testing import get_hips_extra_file, make_test_wcs_geometry, requires_hips_extra
+
+
+def get_test_tiles():
+    # TODO: check if this tile is inside our image
```
cdeil: Resolve this and other TODOs.

hips/draw/tests/test_simple.py
```diff
@@ -1 +1,52 @@
...
+def test_draw_sky_image():
+    geometry = make_test_wcs_geometry(case=2)
+    tiles = get_test_tiles()
+
+    data = draw_sky_image(geometry, tiles)
+
+    assert data.shape == geometry.shape
+    assert data.dtype == np.float64
+    assert np.sum(data) == 4575235421.5126467
+    assert_allclose(data[400, 500:504], [2866.010141, 2563.691673, 2580.759013, 2746.608711])
```
cdeil: Thinking about it some more, I think tests would be a little easier to read and maintain if you do the asserts just one one pixel at a time, instead of multiple pixels in a row.
So here and below, I'd suggest you test two pixels, and do the asserts on separate lines.

hips/draw/tests/test_simple.py
```diff
@@ -1 +1,52 @@
...
+    url = 'https://raw.githubusercontent.com/hipspy/hips-extra/master/datasets/samples/DSS2Red/properties'
+    hips_survey = HipsSurveyProperties.fetch(url)
+    geometry = make_test_wcs_geometry(case=2)
+
+    data = make_sky_image(geometry, hips_survey)
+
+    assert data.shape == geometry.shape
+    assert data.dtype == np.float64
+    assert_allclose(data[200, 994:1000], [3717.10091363, 3402.55292158, 3181.16613051, 2868.45175662, 2832.23001706,
+                                          2779.23366271])
```
cdeil: See my comment above: I don't like the long and broken line, and if the assert fails it will be hard to read which pixel has which value, i.e. what is wrong.
So please just assert on two pixels, and put those as two different assert statements on separate lines.

hips/tiles/tile.py
```diff
@@ -95,7 +95,7 @@ def read(cls, meta: HipsTileMeta, filename: str = None) -> 'HipsTile':
 
         if meta.file_format == 'fits':
             hdu_list = fits.open(str(path))
-            data = hdu_list[0].data
+            data = hdu_list[0].data.astype('float')
```
cdeil: Is this `dtype` conversion here really needed? Is it a good idea?
Naively I would think that we should just process data in the dtype given to us, i.e. remove the `.astype('float')` here.
If dtype conversions are needed, then we should have a logic about it, i.e. please leave a short comment why it's done here.

hips/tiles/tile_meta.py
```diff
@@ -49,3 +52,13 @@ def filename(self) -> str:
...
         """Full path (folder and filename) (`~pathlib.Path`)"""
         return self.path / self.filename
+
+    @property
+    def nside(self):
+        """Return the nside of the HEALPix map"""
+        return hp.order2nside(self.order)
+
+    @property
+    def dst(self):
```
cdeil: Missing docstring.

hips/tiles/tile_meta.py
```diff
@@ -49,3 +52,13 @@ def filename(self) -> str:
...
+        return np.array(
+            [[self.tile_width - 1, 0], [self.tile_width - 1, self.tile_width - 1], [0, self.tile_width - 1], [0, 0]])
```
cdeil: Please wrap this line, i.e. have each corner on a separate line.

hips/tiles/tile.py
adl1995: @cdeil Yes, this is required. Without this, the pixel values are very small for some reason, maybe they get rounded off without the explicit mention of `float`.

cdeil: What was the dtype before the conversion? Print out data.dtype.
Probably changing the data dtype on I/O isn't a good solution (because you want to be able to read and write back tiles unchanged).
So if needed at all this should happen when processing the tile for drawing, no?

In any case, please leave a comment why you're doing the `astype` call on the line above.

adl1995: I just tried this and it prints `>i2`. So, I searched this up and it means 'big-endian byte integer', more [here](https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.byteorder.html).

tboch: DSS2Red FITS tiles are encoded as 16-bits integers (BITPIX = 16), so that makes sense to me.

cdeil: OK, so I'm not surprised if the drawing doesn't work out of the box with integer data.
One has to read the fineprint and experiment / think a bit whether the transform can work on integer data or whether one has to convert to float before.
@adl1995 - See http://scikit-image.org/docs/stable/user_guide/data_types.html

So my suggestion here would be:
- Move the conversion to float to later, e.g. when accessing the tile data for drawing (not on read)
- Add a test here for int and float tiles to establish the current behaviour of our code
- If  you can't figure out all int vs float related conversion issues for drawing by early tomorrow, make a reminder issue to continue working on that in a follow-up pull request next week.
(I really want this PR finished and merged by tomorrow, so that I can start playing with the code easily in master).

hips/tiles/tile.py
```diff
@@ -95,6 +95,7 @@ def read(cls, meta: HipsTileMeta, filename: str = None) -> 'HipsTile':
 
         if meta.file_format == 'fits':
             hdu_list = fits.open(str(path))
+            # This converts a big-endian byte integer to a float
```
cdeil: Remove this comment. It no longer applies.

hips/tiles/tile_meta.py
```diff
@@ -19,12 +27,22 @@ class HipsTileMeta:
...
+    >>> from hips.tiles import HipsTileMeta
+    >>> tile_meta = HipsTileMeta(order=3, ipix=450, file_format='fits', frame='galactic', tile_width=512)
+    >>> tile_meta.skycoord_corners
+    <SkyCoord (Galactic): (l, b) in deg
+    [( 264.375, -24.62431835), ( 258.75 , -30.        ),
+     ( 264.375, -35.68533471), ( 270.   , -30.        )]>
     """
 
-    def __init__(self, order: int, ipix: int, file_format: str, tile_width: int = 512) -> None:
+    def __init__(self, order: int, ipix: int, file_format: str, frame='galactic', tile_width: int = 512) -> None:
```
cdeil: You need to add `frame` to the `Parameters` section in the docstring and document what the valid values are.
Maybe like this?
```
frame : {'icrs', 'galactic'}
    Sky coordinate frame
```

hips/tiles/tile_meta.py
```diff
@@ -37,15 +55,47 @@ def __eq__(self, other: 'HipsTileMeta') -> bool:
...
+            [[self.tile_width - 1, 0],
+             [self.tile_width - 1, self.tile_width - 1],
+             [0, self.tile_width - 1],
+             [0, 0]])
+
+    @property
+    def skycoord_corners(self) -> SkyCoord:
+        """Corner values for a HiPS tile"""
+        theta, phi = boundaries(self.nside, self.ipix)
+        if self.frame == 'galactic':
```
cdeil: If you use the values that SkyCoord support for the `frame` attribute also on this class, then you don't need the if-else here.

hips/tiles/tile_meta.py
```diff
@@ -37,15 +55,47 @@ def __eq__(self, other: 'HipsTileMeta') -> bool:
...
+            return SkyCoord(l=phi, b=np.pi / 2 - theta, unit='radian', frame=self.frame)
+        else:
+            return SkyCoord(ra=phi, dec=np.pi / 2 - theta, unit='radian', frame=self.frame)
+
+    def apply_projection(self, wcs: WCS) -> tf.ProjectiveTransform:
```
cdeil: I know it's still very much unclear which code goes where.
But I would strongly suggest to not make this a property on `HipsTileMeta`, but leave it on the painter.

hips/utils/testing.py
```diff
@@ -0,0 +1,55 @@
...
+    """Decorator to mark tests requiring ``hips-extra`` data.
+    """
+    has_hips_extra = get_hips_extra_file('datasets/samples/DSS2Red/properties').is_file()
+    skip_it = not has_hips_extra
+    reason = 'No hips-extra data available.'
+    return pytest.mark.skipif(skip_it, reason=reason)
+
+
+def make_test_wcs_geometry(case=0):
+    """An example Galactic CAR WCS that """
```
cdeil: This docstring ends abruptly (and is incorrect, says CAR but there's an AIT case below.
Maybe just remove it?

hips/draw/tests/test_simple.py
```diff
@@ -1 +1,71 @@
...
+        geometry = make_test_wcs_geometry(case=2)
+        tile = HipsTile.read(
+            meta=HipsTileMeta(order=3, ipix=450, file_format='fits', frame='icrs', tile_width=512),
+            filename=get_hips_extra_file('datasets/samples/DSS2Red/Norder3/Dir0/Npix450.fits'),
+        )
+        cls.simple_tile_painter = SimpleTilePainter(geometry, tile)
+
+    def test_warp_image(self):
+        self.simple_tile_painter.warp_image()
+        assert_allclose(self.simple_tile_painter.tile.meta.skycoord_corners.ra.deg,
```
cdeil: Since `skycoord_corners` is a property on `HipsTileMeta`, this assert belongs in a test for that class (not this painter class).

hips/draw/simple.py
```diff
@@ -1,5 +1,111 @@
...
+    """
+
+    def __init__(self, geometry: WCSGeometry, tile: HipsTile) -> None:
+        self.geometry = geometry
+        self.tile = tile
+
+    def warp_image(self) -> np.ndarray:
+        """Warp a HiPS tile and a sky image"""
+        pt = self.tile.meta.apply_projection(self.geometry.wcs)
+        return tf.warp(self.tile.data.astype('float'), pt, output_shape=self.geometry.shape)
```
cdeil: Move the type conversion on a separate line.
Also note that http://scikit-image.org/docs/dev/user_guide/data_types.html says "never use astype" when using scikit-image. Please read that page and see if there's a better way.
In any case, please leave a comment saying why your changing the dtype here.

hips/tiles/tile_meta.py
adl1995: Could you please elaborate? I tried this:

`return SkyCoord(phi, np.pi / 2 - theta, unit='radian', frame=self.frame)`

But, it doesn't give the correct results.

cdeil: Why not? What to you pass for "frame"?
Maybe print out the inputs and outputs of the function you have now, and with the change I proposed, and then you should understand what's going on?

hips/draw/simple.py
adl1995: I just tried:

```
from skimage import img_as_float
img_as_float(self.tile.data)
```

But, it produces very small values in the range `0.08748028201677939` as opposed to `2866.0101409848185` when using `astype`.

hips/tiles/tile_meta.py
```diff
@@ -17,14 +23,26 @@ class HipsTileMeta:
...
+    >>> from hips.tiles import HipsTileMeta
+    >>> tile_meta = HipsTileMeta(order=3, ipix=450, file_format='fits', frame='galactic', tile_width=512)
+    >>> tile_meta.skycoord_corners
+    <SkyCoord (Galactic): (l, b) in deg
+    [( 264.375, -24.62431835), ( 258.75 , -30.        ),
+     ( 264.375, -35.68533471), ( 270.   , -30.        )]>
     """
 
-    def __init__(self, order: int, ipix: int, file_format: str, tile_width: int = 512) -> None:
+    def __init__(self, order: int, ipix: int, file_format: str, frame: str = 'galactic', tile_width: int = 512) -> None:
```
cdeil: I suggest to change the default to "icrs". That's more common for HiPS data, and it's also the default for `SkyCoord`.

hips/draw/simple.py
```diff
@@ -1,5 +1,122 @@
...
+    tiles : List[HipsTile]
+        A list of HipsTile
+
+    Returns
+    -------
+    np.ndarray
+        Returns a numpy array containing all HiPS tiles projected onto it
+    """
+    # TODO: Fix type annotation issue
+    all_sky = np.zeros(geometry.shape)
```
cdeil: Usually the image won't be all of the sky, just a small part. So I'd suggest to rename "all_sky" to something else here. E.g. "sky_image" or just "image" or "data".

hips/draw/simple.py
```diff
@@ -1,5 +1,122 @@
...
+    for tile in tiles:
+        draw_tile = SimpleTilePainter(geometry, tile)
```
cdeil: I find it confusing if you use the name "draw_tile" for the painter, because it's not a "tile" object.
Suggest to rename to "painter" here.

hips/draw/tests/test_simple.py
```diff
@@ -1 +1,52 @@
...
+from astropy.utils.data import get_pkg_data_filename
+from numpy.testing import assert_allclose
+
+from ..simple import make_sky_image, draw_sky_image
+from ...tiles import HipsSurveyProperties, HipsTileMeta, HipsTile
+from ...utils.testing import get_hips_extra_file, make_test_wcs_geometry, requires_hips_extra
+
+
+def get_test_tiles():
+    frames = dict({'equatorial': 'icrs', 'galactic': 'galactic', 'ecliptic': 'ecliptic'})
```
cdeil: This mapping between HiPS to SkyCoord strings for frame shouldn't be here in the test.
It should be available somewhere else, in a central place, and then if you need to call it from here (or just hard-code the values you need here in `get_test_tiles`.
Maybe putting the dict on the `HipsTileMeta` class would be a good location? Or somewhere in `hips.utils`?

cdeil:
At the end of the tests there's these warnings:
https://travis-ci.org/hipspy/hips/jobs/248660210#L2173
```
/private/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/hips-test-0qyifajc/lib.macosx-10.9-x86_64-3.6/hips/draw/tests/test_simple.py:24: ResourceWarning: unclosed file <_io.FileIO name='/Users/travis/hips-extra/datasets/samples/DSS2Red/Norder3/Dir0/Npix451.fits' mode='rb' closefd=True>
```
The reason is that your `fits.open` calls leave open file handles. Please change the FITS reading in `HipsTile.read` to this pattern, that should (I hope) resolve the issue, i.e. close the file handles at the end of the with block, before exiting the read method call:
```python
with fits.open(str(path) as hdu_list:
        data = hdu_list[0].data
        header = hdu_list[0].header
return cls(meta, data, header)
```

cdeil:
Please make sure all tests are passing (or ask me for help if you don't know how to resolve some fails). See https://travis-ci.org/hipspy/hips/jobs/248660210#L2135

cdeil:
Do you know what's up with this warning?
https://travis-ci.org/hipspy/hips/jobs/248660210#L2161
```
/Users/travis/miniconda/envs/test/lib/python3.6/site-packages/astropy/io/fits/file.py:336: AstropyUserWarning: File may have been truncated: actual file length (527168) is smaller than the expected size (529920)
    self.size, pos), AstropyUserWarning)
```
Can you produce a few lines with a standalone test case that results in the same issue so that we can understand / debug / fix the issue?

hips/utils/testing.py
```diff
@@ -0,0 +1,54 @@
...
+    and set an environment variable ``HIPS_EXTRA`` pointing to it.
+    """
+    path = Path(os.environ['HIPS_EXTRA'])
+    return path / filename
+
+
+def requires_hips_extra():
+    """Decorator to mark tests requiring ``hips-extra`` data.
+    """
+    has_hips_extra = get_hips_extra_file('datasets/samples/DSS2Red/properties').is_file()
```
cdeil: Locally I didn't have `HIPS_EXTRA` set, and then the tests fail like this:
```
/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/os.py:666: in __getitem__
    value = self._data[self.encodekey(key)]
E   KeyError: b'HIPS_EXTRA'

During handling of the above exception, another exception occurred:
hips/draw/tests/test_simple.py:30: in <module>
    @requires_hips_extra()
hips/utils/testing.py:28: in requires_hips_extra
    has_hips_extra = get_hips_extra_file('datasets/samples/DSS2Red/properties').is_file()
hips/utils/testing.py:21: in get_hips_extra_file
    path = Path(os.environ['HIPS_EXTRA'])
/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/os.py:669: in __getitem__
    raise KeyError(key) from None
E   KeyError: 'HIPS_EXTRA'
_____________________________________________________________________ ERROR collecting lib.macosx-10.12-x86_64-3.6/hips/draw/tests/test_simple.py ______________________________________________________________________
/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/os.py:666: in __getitem__
    value = self._data[self.encodekey(key)]
E   KeyError: b'HIPS_EXTRA'

During handling of the above exception, another exception occurred:
hips/draw/tests/test_simple.py:30: in <module>
    @requires_hips_extra()
hips/utils/testing.py:28: in requires_hips_extra
    has_hips_extra = get_hips_extra_file('datasets/samples/DSS2Red/properties').is_file()
hips/utils/testing.py:21: in get_hips_extra_file
    path = Path(os.environ['HIPS_EXTRA'])
/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/os.py:669: in __getitem__
    raise KeyError(key) from None
E   KeyError: 'HIPS_EXTRA'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=============================================================================================== 2 error in 0.40 seconds ================================================================================================
```
The reason is simply that the `requires_hips_extra` implementation is buggy.
This should work:
```
def has_hips_extra():
    """Is hips-extra available? (bool)"""
    if 'HIPS_EXTRA' in os.environ:
        path = Path(os.environ['HIPS_EXTRA']) / 'datasets/samples/DSS2Red/properties'
        if path.is_file():
            return True

    return False


def requires_hips_extra():
    """Decorator to mark tests requiring ``hips-extra`` data.
    """
    skip_it = not has_hips_extra()
    reason = 'No hips-extra data available.'
    return pytest.mark.skipif(skip_it, reason=reason)
```
I didn't test that locally. @adl1995 - Please put it and you test locally that it works as expected by running the tests with and without having `HIPS_EXTRA` env variable set.

hips/tiles/tile_meta.py
adl1995: I passed galactic for frame. I think we need to explicitly mentioned, `l`, `b`, `ra`, or `dec`. Otherwise, it assumes different parameters.

adl1995:
@cdeil I made another series of commits. Currently waiting to Travis CI build to complete.

adl1995:
@cdeil I'm not sure why this example is failing to execute (https://travis-ci.org/hipspy/hips/jobs/248711430#L1182). I have defined `hips_survey` in the above line, but it still raises a `NameError`.

hips/draw/simple.py
```diff
@@ -1,5 +1,122 @@
...
+    geometry : `~hips.utils.WCSGeometry`
+        An object of WCSGeometry
+    tile : `HipsTile`
+       An object of HipsTile
+    """
+
+    def __init__(self, geometry: WCSGeometry, tile: HipsTile) -> None:
+        self.geometry = geometry
+        self.tile = tile
+        self.pt = None
```
cdeil: Remove `self.pt` here, and also in the property below, simply use a local variable `pt`, don't store it on `self`.

hips/draw/simple.py
```diff
@@ -1,5 +1,121 @@
...
+        corners = self.tile.meta.skycoord_corners.to_pixel(self.geometry.wcs)
+        src = np.array(corners).T.reshape((4, 2))
+        dst = self.tile.meta.dst
+        pt = tf.ProjectiveTransform()
+        pt.estimate(src, dst)
+        return pt
+
+    def warp_image(self) -> np.ndarray:
+        """Warp a HiPS tile and a sky image"""
+        return tf.warp(self.tile.data, self.projection, output_shape=self.geometry.shape, preserve_range=True)
```
cdeil: Wrap this line.

docs/getting_started.rst
```diff
@@ -1,9 +1,31 @@
 .. include:: references.txt
 
 .. _gs:
+.. doctest-skip-all
```
cdeil: does linking to the `gs` label still work if you put the `.. doctest-skip-all` between it and the heading?
Even if yes, I'd suggest to exchange the order of those two lines.

docs/getting_started.rst
```diff
@@ -1,9 +1,31 @@
...
+    from hips.utils import WCSGeometry
+    from hips.draw import make_sky_image
+    from hips.tiles import HipsSurveyProperties
+    geometry = WCSGeometry.create(
+         skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
+         shape=(1000, 2000), coordsys='GAL',
+         projection='AIT', cdelt=0.01, crpix=(1000, 500),
+    )
+    url = 'https://raw.githubusercontent.com/hipspy/hips-extra/master/datasets/samples/DSS2Red/properties'
+    hips_survey = HipsSurveyProperties.fetch(url)  # doctest: +REMOTE_DATA
```
cdeil: Suggest to remove the `# doctest: +REMOTE_DATA` here.

cdeil:
Travis-ci is all green now: https://travis-ci.org/hipspy/hips/builds/248735077?utm_source=github_status&utm_medium=notification

@adl1995 - Merge now? Or do you want to continue today and try to fix the computation of the list of tiles needed here to take into account the Hips and sky image coordinate frame correctly?

cdeil:
After some more debugging and pair-coding with @adl1995 we have a working `make_sky_image`, see the attached screenshot of the image resulting from the high-level docs example.

Well, it's kinda working, you can see that the summing of the tiles creates bright line artefacts at the edges, and that tiles are missing because they query_disk doesn't work properly.

Let's discuss those and other things in separate follow-up issues. I'm merging this now.

@adl1995 - Good job! And have a nice weekend!

screen shot 2017-06-30 at 16 57 06

tboch:
Well done, that's a valuable progress!

PR #28, opened 2017-06-23
JPEG I/O library

cdeil:
In #24 is using PIL/Pillow to read JPEG tiles. This is a reminder issue that we need to declare and document this dependency (in `setup.py`, `travis.yml`, the package install instructions).


cdeil:
I've attached a commit here adding Pillow as a dependency.

PR #33, opened 2017-06-25
Outline first draft of HipsTileCache class

adl1995:
A work in progress of `HipsTileCache` class.

### DIFF
--- a/hips/tiles/__init__.py
+++ b/hips/tiles/__init__.py
@@ -4,3 +4,4 @@ from .tile import *
 from .tile_meta import *
 from .description import *
 from .surveys import *
+from .cache import *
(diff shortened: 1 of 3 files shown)


cdeil:
Suggest to rebase this PR against master and then to start development here by moving the existing `hips.draw.fetch_tiles` over to the `hips.tile` sub-package, e.g. the file `hips/tile/cache.py` you're adding here.

cdeil:
I'm moving this to the v0.2 milestone. I think just what we have now is very close to v0.1 and having a cache can be in the second version.

cdeil:
@adl1995 doesn't plan to continue with this, but we're leaving it open just as a reminder / possible starting point if someone picks this up in the future.

cdeil:
Discussed with @adl1995  just now. Decided to close this PR that is 2 years old. We're note sure what the idea / use case here was. Probably to support cloning part of HiPS from a server to a local disk. For this the feature request #122 is still open.

Issue #17, opened 2017-06-03
Function to compute WCS for a given FITS HIPS tile


cdeil:
@tboch - I see you gave an example of a FITS WCS for a given FITS HIPS tile:
https://hips.readthedocs.io/en/latest/drawing_algo.html#wcs-for-fits-tiles

Can you describe how it could be computed for any tile?

Maybe @adl1995 (or you if it's too complex) could add a function (e.g. in `hips/utils.py` or a new module) that returns such an `astropy.wcs.WCS` for given HiPS tile parameters?

I think that would be valuable to have, both for debugging and precise tile drawing (e.g. `reproject` just calls into `wcslib` via `astropy.wcs.WCS`, so probably reprojecting a tile with WCS into any other sky image will just work with one call into `reproject` (and they have the slow flux conserving method also implemented, which is nice).

tboch:
My colleague François-Xavier provided me with a Java executable to compute this WCS.
I know he planned to release his HEALPix-related code under a LGPL licence, I will check if he can already provide me the source code.

cdeil:
Moving to v0.2 milestone, this isn't urgent. Still, @tboch - any news?

tboch:
I have the Java source code available. I'll provide a Python version by end of this week.
Where should I put this?

cdeil:
If you have time, you could add this directly to the `hips` package.
Maybe a property or method on https://hips.readthedocs.io/en/latest/api/hips.HipsTileMeta.html would be a good location?

If you want @adl1995 to add it (and add tests / docstring), then you could just put the code you have in a  https://gist.github.com/ and then @adl1995 will make a PR adding everything.

cdeil:
@tboch - Coming back to this: is it really possible to have a WCS for a HiPS tile?

If yes, I think this would be very nice. E.g. one could just use http://reproject.readthedocs.io/ to get a precise tile painting or tile generation if going the other way. And also for debugging, it's very nice to be able to add a WCS and e.g. open a tile in DS9 or check world and pixel coords from Python.

tboch:
@cdeil : yes, it's possible, I have some working code I still have to properly validate.
Should I create a different branch for this?

cdeil:
Yes, always work in a git branch.

Probably it's a function that makes a WCS or a Header object?
Could go here: https://github.com/hipspy/hips/tree/master/hips/utils

If you have an implementation, validating shouldn't be too hard.
[...]

PR #37, opened 2017-06-26
Add WCS property in HipsTile class

adl1995:
I added a `WCS` property in `HipsTile` class. As drawing is going to be a major part of this library, I think this is a useful addition to the class.

### DIFF
--- a/hips/tiles/tests/test_tile.py
+++ b/hips/tiles/tests/test_tile.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-from numpy.testing import assert_equal
 from astropy.tests.helper import remote_data
+from numpy.testing import assert_equal
+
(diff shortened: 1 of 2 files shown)


cdeil:
@adl1995 - I don't think this will work. Currently the FITS headers of HiPS tiles in FITS format don't contain a valid WCS that would be useful for drawing. See #17.

adl1995:
But, wouldn't we still require this later on?

cdeil:
@adl1995 - Not necessarily, we can draw the tile without having a WCS for the tile itself.

It would be nice to have though, since if we have a WCS for the tile we can just call reproject as an alternative precise drawing method, as mentioned in #17.

There's nothing you can do here for now, I'd suggest to just close this PR without merging.
Maybe in the future, Thomas will provide code to compute the tile WCS, or maybe we won't go that way.

Issue #36, opened 2017-06-26
How to handle tiles in different formats?


cdeil:
Currently we have a class [HipsTile](https://hips.readthedocs.io/en/latest/api/hips.HipsTile.html) with data members:
* meta - https://hips.readthedocs.io/en/latest/api/hips.HipsTileMeta.html (containing "format" string)
* data - numpy array
* header - http://docs.astropy.org/en/stable/io/fits/api/headers.html#astropy.io.fits.Header

We want to support the three formats used for HiPS: fits, jpeg, png

What we have now isn't nice, because `header` is a FITS header and doesn't apply to the jpeg and png case. But more importantly, as mentioned in #35, our current scheme can't handle JPEG well, because it's a lossy format, so to / from Numpy array can change the pixel data.

One feature we need is to be able to download and store on local disk JPEG tiles, unaltered, i.e. not going though JPEG decompression to Numpy array and compression again to JPEG, resulting in different pixel data. This means we have to store the JPEG file contents (i.e. `raw_image = BytesIO(urllib.request.urlopen(url).read())` as a data member on the JPEG HiPS tile, so that we can write it unaltered to local disk or cache if we want.

There's a few options:
1. add a `.raw_data` member to `HipsTile` and change `.data` into a property (maybe rename to `.array_data`) that's computed from `.raw_data` when needed (i.e. for drawing).
2. Introduce a class hierarchy `HipsTileJpeg`, `HipsTilePng`, `HipsTileFits` with `HipsTile` as an abstract base class. We would still add `.raw_data` at least for `HipsTileJpeg`, it would mostly be a code re-organisation where the `if-else` stuff that you currently do for each `format` in `HipsTile` gets split out into separate classes / methods.
3. An alternative would be to introduce a `HipsTileData` base class and then do the `HipsTileDataJpeg` etc differences only on these data classes, with a single common `HipsTile` class. The question if this or option 2 works better has to do with the question how much the data and metadata/header handling is coupled. I think we discussed this previously and said that for JPEG and PNG tiles, the file header doesn't contain any useful information and is irrelevant.

I think overall this is a difficult question, but one that we can mostly punt on for now and only come back to if / when we start generating HiPS tiles, for now we only need 1:1 copy and reading. So I think my suggestion for now would be to just do the small change outlined as option 1, i.e. to keep `rawdata` around at least for the JPEG case and to use that from `write`.

@adl1995 @tboch - Thoughts?

adl1995:
[...]

PR #38, opened 2017-06-28
Rename data to raw_data

adl1995:
This PR addresses issue #36.

### DIFF
--- a/hips/tiles/tile.py
+++ b/hips/tiles/tile.py
@@ -1,11 +1,13 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-from pathlib import Path
 import urllib.request
 from io import BytesIO
+from pathlib import Path
+
 import numpy as np
 from PIL import Image
 from astropy.io import fits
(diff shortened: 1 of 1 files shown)


cdeil:
@adl1995 - As-is, this isn't really useful. You're just doing a rename `data` -> `raw_data`, and then adding a property `data` that aliases `raw_data`.

My suggestion in #36 was to **add** a new property called `raw_data` that stores the bytes from tile files, i.e. something like:
```
self.raw_data = BytesIO(urllib.request.urlopen(url).read())
```
and to make a property `@data` that extracts the Numpy array from the bytes.

@adl1995 - If it's still not clear how this should work, please put it on the agenda for the call today and I'll explain.

cdeil:
@adl1995 - I'm closing this old PR.

As I said in my last comment and #36 , it doesn't properly implement what we need, there's already a merge conflict and there will be more merge conflicts after #46 .

My suggestion is to start a new branch / PR next week after #46 is in.

PR #25, opened 2017-06-30
Move hips package from Python 3.5 to 3.6

cdeil:
Like I've mentioned in a note at https://hips.readthedocs.io/en/latest/installation.html I would suggest to only support Python 3.6+ for the hips package, i.e. to drop Python 3.5 support.

@tboch or anyone - if you think it's a bad idea and we should support Python 3.5, please comment.
This is more of a "nice to be able to use the latest features as a developer", there is nothing in Python 3.6 that we absolutely need, we can support 3.5.

For now, I'm holding off on this change because: 
- healpy packages for Python 3.6 are missing on conda-forge, i.e. we can't switch travis-ci to Python 3.6 yet.
- readthedocs doesn't support Python 3.6 yet (at least they didn't 2 weeks ago, should be resolved soon).

So this is a reminder issue for myself, for now assigned to the v0.2 milestone.

### DIFF
--- a/.rtd-environment.yml
+++ b/.rtd-environment.yml
@@ -5,9 +5,7 @@ channels:
     - astropy
 
 dependencies:
-    # TODO: change to Python 3.6 once it's supported:
-    # https://github.com/rtfd/readthedocs.org/issues/2584#issuecomment-304366679
-    - python=3.5
+    - python=3.6
     - numpy
     - astropy
     - healpy
--- a/hips/utils/wcs.py
+++ b/hips/utils/wcs.py
@@ -102,13 +102,13 @@ class WCSGeometry:
         w = WCS(naxis=2)
 
         if coordsys == 'CEL':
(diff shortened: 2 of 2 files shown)


cdeil:
I've attached a commit here that uses f-strings, a feature added in Python 3.6, i.e. drops Python 3.5 compatibility. travis-ci was already changed to run on Python 3.6, so I expect tests to pass.

cdeil:
Looks like the change Python 3.5 to 3.6 on ReadtheDocs via this commit in master worked:
* https://github.com/hipspy/hips/commit/00e9dea625a6d9bf19fa21a13a606531c6da7279
* https://readthedocs.org/projects/hips/builds/5629248/

cdeil:
I've cherry-picked 4baa6434a87ab754167be1b32006cff6a1070720 into master.
Closing this issue now.

Issue #42, opened 2017-06-30
HIPS FITS tiles truncated?


cdeil:
When running the tests, at the end warnings like these appear:
```
WARNING: File may have been truncated: actual file length (527168) is smaller than the expected size (529920) [astropy.io.fits.file]
```
This has nothing to do with our hips package, I think the FITS tiles in hips-extra might simply have some issue:
```
In [1]: from astropy.io import fits

In [2]: fits.open('datasets/samples/IRAC4/Norder3/Dir0/Npix307.fits')
WARNING: File may have been truncated: actual file length (1051456) is smaller than the expected size (1054080) [astropy.io.fits.file]
Out[2]: [<astropy.io.fits.hdu.image.PrimaryHDU object at 0x109d6aef0>]
```

```
$ ftverify datasets/samples/IRAC4/Norder3/Dir0/Npix307.fits
 
               ftverify 4.17 (CFITSIO V3.370)               
               ------------------------------               
 
HEASARC conventions are being checked.
 
File: datasets/samples/IRAC4/Norder3/Dir0/Npix307.fits

1 Header-Data Units in this file.
 
=================== HDU 1: Primary Array ===================
 
*** Error:   checking data fill: error reading from FITS file
 
 8 header keywords
 
 32-bit floating point pixels,  2 axes (512 x 512), 
 
++++++++++++++++++++++ Error Summary  ++++++++++++++++++++++
 
 HDU#  Name (version)       Type             Warnings  Errors
 1                          Primary Array    0         1     
 
**** Verification found 0 warning(s) and 1 error(s). ****
```

@tboch - Can you please have a look?

tboch:
Yes indeed, FITS tiles we generated some time ago are not FITS-valid as they are missing the final bytes (the total size is not a multiple of 2880).
This generates some warnings when reading them, but actual data is fine.

Unless you disagree, I think we can close this one

cdeil:
I would suggest this: #46

cdeil:
Resolved in #46 by suppressing the warnings on FITS tile read.

PR #46, opened 2017-07-02
Improve HiPS tile I/O

cdeil:
I would like to propose this solution to address #42 : suppress the warnings from invalid FITS tiles.

Since this is a "known issue" (or "known feature" if you like) for the FITS tiles from CDS, and every user will get one of these warnings per tile they fetch or read on the console, and there's nothing they can / should do about it except to ignore it, I think it is appropriate to suppress these warnings.

I'm not an expert on Python warnings, but I think / hope that this will have little other effect than suppressing the warnings we want to suppress, and to keep other warnings / errors intact.

@tboch - Assigning to you for review.

@adl1995 - I've also cleaned up the I/O functions a bit to:
* always use `with` statements to read / fetch, to avoid leaving open filehandles around
* removed the code duplication in FITS tile read and fetch by always reading into a `BytesIO` and then doing the "decoding" in a `_from_raw_data` method.
* Added PNG as a third case and put NotImplementedError for now, and raise ValueError if not one of the three formats we can handle. I don't know yet if the code to handle JPG and PNG will be identical or different, if yes we could combine these two cases again via `elif file_format in {'jpg', 'png'}`, keeping the `else ValueError` part to give a better error message in case a format we don't support appears.

### DIFF
--- a/hips/tiles/surveys.py
+++ b/hips/tiles/surveys.py
@@ -65,8 +65,8 @@ class HipsSurveyProperties:
             URL containing HiPS properties
         """
 
-        response = urllib.request.urlopen(url).read()
-        text = response.decode('utf-8')
+        with urllib.request.urlopen(url) as response:
+            text = response.read().decode('utf-8')
         return cls.parse(text)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


cdeil:
@tboch - Can you comment on this PR (or merge it)?

@adl1995 wants to add support / tests for PNG tiles this week, and it's easiest if this PR (or parts of it) go in first.

cdeil:
@adl1995 - Now that this PR is in, you could do #38 (i.e. fully support JPEG read / write, making `raw_data` a `HipsTile` data member of type `BytesIO`) or implement code and tests for PNG tiles.

Note that the current tile I/O tests combine fetch / read / write. Now that we have `@requires_hips_extra` I would suggest to use `hips-extra` for read / write and all tile tests, and to only have one small test for the fetching using `@remote_data`.

Issue #39, opened 2017-06-29
Add WCSGeometry create_simple that matches Aladin Lite


cdeil:
Currently we have this convenience methods that users can use to create a WCS:
https://hips.readthedocs.io/en/latest/api/hips.utils.WCSGeometry.html#hips.utils.WCSGeometry.create
It's a bit cumbersome to use (e.g. shape and crpix have different order x/y, one has to compute crpix manually before calling).

I think it would be nice to have another one that matches what Aladin Lite gives:
(See http://aladin.u-strasbg.fr/AladinLite/doc/API/#init)

I.e. something like this on `WCSGeometry`
```
@classmethod
def create_simple(cls, skydir: SkyCoord, fov: Angle, shape: tuple of float):
    ... compute WCS as Aladin LIte does it, maybe calling the other create method, maybe not.
    return cls(wcs, shape)
```
@tboch - Could you please make a pull request, or give infos here for @adl1995 how exactly Aladin Lite computes the WCS so that he can implement it?

tboch:
What Aladin Lite uses is similar to this code:

```
def make_wcs(ra, dec, fov_deg, width, height):
    """Create a WCS from a few params
    Parameters
    ----------
    ra : right ascension of the center
    dec : declination of the center
    fov_deg : field of view in degrees
    width : image width in pixels
    height : image height in pixels
    """"    

   resolution = float(fov_deg) / float(max(width, height))

    w = wcs.WCS(naxis=2)
    w.wcs.crpix = [width / 2., height / 2.]
    w.wcs.cdelt = numpy.array([-resolution, resolution])
    w.wcs.crval = [ra, dec]
    # this should be adapted to other type of projections
    w.wcs.ctype = ["RA---SIN", "DEC--SIN"]

    return w
```

Some comments:
 * In Aladin Lite, this has only been tested for SINUS projections but I'd expect it to work for others.
 * the input parameters should of course be adapted to what @cdeil is suggesting
 * we are missing a rotation angle parameter, but that's not needed for a v0.1

cdeil:
@tboch - Thanks!

Note that this is just an "extra constructor" as a convenience. Users still can pass any `WCS` to `__init__` and have the full power of choosing any FITS WCS. So in my opinion we don't need to try and expose everything via these convenience constructors. But yes, where easily possible, we can add options such as rotation or projection string as you mention)

[...]

PR #49, opened 2017-07-03
Add classmethod create_simple to WCSGeometry

adl1995:
Added a classmethod as mentioned in #39.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -20,10 +20,10 @@ To make a sky image with the `hips` package, follow the following three steps:
     from astropy.coordinates import SkyCoord
     from hips.utils import WCSGeometry
 
-    geometry = WCSGeometry.create(
+    geometry = WCSGeometry.create_simple(
          skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
-         shape=(1000, 2000), coordsys='GAL',
-         projection='AIT', cdelt=0.01, crpix=(1000, 500),
+         width=2000, height=1000, fov="3 deg",
+         coordsys='galactic', projection='AIT'
     )
 
 
--- a/hips/utils/testing.py
+++ b/hips/utils/testing.py
@@ -28,7 +28,6 @@ def has_hips_extra():
         path = Path(os.environ['HIPS_EXTRA']) / 'datasets/samples/DSS2Red/properties'
         if path.is_file():
             return True
-
     return False
 
 
@@ -43,19 +42,19 @@ def make_test_wcs_geometry(case=0):
     if case == 0:
         return WCSGeometry.create(
             skydir=SkyCoord(3, 4, unit='deg', frame='galactic'),
-            shape=(2, 3), coordsys='GAL',
+            width=2, height=3, coordsys='galactic',
             projection='CAR', cdelt=1.0, crpix=(1, 1),
         )
     elif case == 1:
         return WCSGeometry.create(
             skydir=SkyCoord(10, 20, unit='deg', frame='galactic'),
-            shape=(10, 20), coordsys='GAL',
+            width=20, height=10, coordsys='galactic',
             projection='CAR', cdelt=1.0, crpix=(1, 1),
         )
     elif case == 2:
         return WCSGeometry.create(
             skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
-            shape=(1000, 2000), coordsys='GAL',
+            width=2000, height=1000, coordsys='galactic',
             projection='AIT', cdelt=0.01, crpix=(1000, 500),
         )
     else:
--- a/hips/utils/tests/test_wcs.py
+++ b/hips/utils/tests/test_wcs.py
@@ -3,6 +3,7 @@ from numpy.testing import assert_allclose
 from astropy.coordinates import SkyCoord
 from ..wcs import WCSGeometry
 
+
 class TestWCSGeometry:
     def setup(self):
         self.wcs_geometry = WCSGeometry.create_simple(
@@ -33,7 +34,3 @@ class TestWCSGeometry:
         assert_allclose(c.dec.deg, 0.00075, atol=1e-2)
         assert_allclose(self.wcs_geometry.wcs.wcs.crpix, [1000., 500.])
         assert_allclose(self.wcs_geometry.wcs.wcs.cdelt, [-0.0015, 0.0015])
-
-    def skycoord_separation(self):
-        skydir=SkyCoord(0, 0, unit='deg', frame='icrs')
-        skydir.separation(skydir).degree == 2
--- a/hips/utils/tests/test_wcs.py
+++ b/hips/utils/tests/test_wcs.py
@@ -1,18 +1,39 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 from numpy.testing import assert_allclose
-from ..testing import make_test_wcs_geometry
-
+from astropy.coordinates import SkyCoord
+from ..wcs import WCSGeometry
 
 class TestWCSGeometry:
     def setup(self):
-        self.wcs_geometry = make_test_wcs_geometry(case=0)
+        self.wcs_geometry = WCSGeometry.create_simple(
+            skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
+            width=2000, height=1000, fov="3 deg",
+            coordsys='galactic', projection='AIT'
+        )
 
-    def test_center(self):
+    def test_galactic_frame_center(self):
         c = self.wcs_geometry.center_skycoord
+
         assert c.frame.name == 'galactic'
-        assert_allclose(c.l.deg, 2, atol=1e-2)
-        assert_allclose(c.b.deg, 4.5, atol=1e-2)
+        assert_allclose(c.l.deg, 359.99, atol=1e-2)
+        assert_allclose(c.b.deg, 0.00075, atol=1e-2)
+        assert_allclose(self.wcs_geometry.wcs.wcs.crpix, [1000., 500.])
+        assert_allclose(self.wcs_geometry.wcs.wcs.cdelt, [-0.0015, 0.0015])
 
     def test_celestial_frame(self):
-        pass
-        # TODO (Adeel): add a test
+        wcs_geometry = WCSGeometry.create_simple(
+            skydir=SkyCoord(0, 0, unit='deg', frame='icrs'),
+            width=2000, height=1000, fov="3 deg",
+            coordsys='icrs', projection='AIT'
+        )
+        c = wcs_geometry.center_skycoord
+
+        assert c.frame.name == 'icrs'
+        assert_allclose(c.ra.deg, 359.99, atol=1e-2)
+        assert_allclose(c.dec.deg, 0.00075, atol=1e-2)
+        assert_allclose(self.wcs_geometry.wcs.wcs.crpix, [1000., 500.])
+        assert_allclose(self.wcs_geometry.wcs.wcs.cdelt, [-0.0015, 0.0015])
+
+    def skycoord_separation(self):
+        skydir=SkyCoord(0, 0, unit='deg', frame='icrs')
+        skydir.separation(skydir).degree == 2
--- a/hips/utils/wcs.py
+++ b/hips/utils/wcs.py
@@ -58,7 +58,7 @@ class WCSGeometry:
 
     def __init__(self, wcs: WCS, width: int, height: int) -> None:
         self.wcs = wcs
-        self.shape = Shape(*(width, height))
+        self.shape = Shape(width=width, height=height)
 
     @property
     def center_pix(self) -> Tuple[float, float]:
@@ -177,7 +177,7 @@ class WCSGeometry:
         Shape(width=2000, height=1000)
         """
         fov = Angle(fov)
-        crpix = (float(width / 2), float(height / 2))
+        crpix = float(width / 2), float(height / 2)
         cdelt = float(fov.degree) / float(max(width, height))
 
         return cls.create(skydir, width, height, coordsys, projection, cdelt, crpix)
--- a/hips/utils/wcs.py
+++ b/hips/utils/wcs.py
@@ -4,16 +4,19 @@ from typing import Tuple
 import numpy as np
 from astropy.coordinates import SkyCoord
 from astropy.wcs import WCS
+from astropy.coordinates import Angle
 from astropy.wcs.utils import pixel_to_skycoord, wcs_to_celestial_frame
 
 __all__ = [
     'WCSGeometry',
 ]
 
-__doctest_skip__ = ['WCSGeometry']
+__doctest_skip__ = [
+    '*',
+]
 
-Shape = namedtuple('Shape', ['ny', 'nx'])
-"""Helper for 2-dim image shape, to make it clearer what value is x and y."""
+Shape = namedtuple('Shape', ['width', 'height'])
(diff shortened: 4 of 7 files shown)


hips/utils/wcs.py
```diff
@@ -124,6 +124,30 @@ def create(cls, skydir: SkyCoord, shape: tuple, coordsys: str = 'CEL',
...
+        shape : `tuple`
+            Shape of the image (Numpy axis order: y, x)
+        fov_deg : `float`
+            Field of view in degrees
+        coordsys : `str`
+            Coordinate system
+        projection : `str`
+            Projection of the WCS object
+        """
+
```
cdeil: Remove empty line.

hips/utils/wcs.py
```diff
@@ -124,6 +124,30 @@ def create(cls, skydir: SkyCoord, shape: tuple, coordsys: str = 'CEL',
 
         return cls(w, shape)
 
+    @classmethod
+    def create_simple(cls, skydir: SkyCoord, shape: tuple, fov_deg: float,
+                      coordsys: str = 'CEL', projection: str = 'AIT') -> 'WCSGeometry':
```
cdeil: For `coordsys`, I would suggest we change this method and `create` to support `{'icrs', 'galactic'}`.
Having yet another code in the end-user API for the RADEC frame ("CEL" at the moment) is confusing.

hips/utils/wcs.py
```diff
@@ -124,6 +124,30 @@ def create(cls, skydir: SkyCoord, shape: tuple, coordsys: str = 'CEL',
...
+        ----------
+        skydir : `~astropy.coordinates.SkyCoord`
+            Sky coordinate of the WCS reference point
+        shape : `tuple`
+            Shape of the image (Numpy axis order: y, x)
+        fov_deg : `float`
+            Field of view in degrees
+        coordsys : `str`
+            Coordinate system
+        projection : `str`
```
cdeil: You should link to http://docs.astropy.org/en/stable/wcs/#supported-projections so that users know what choices are available.

hips/utils/wcs.py
```diff
@@ -124,6 +124,30 @@ def create(cls, skydir: SkyCoord, shape: tuple, coordsys: str = 'CEL',
...
+                      coordsys: str = 'CEL', projection: str = 'AIT') -> 'WCSGeometry':
+        """Create WCS object programmatically using field of view (`WCSGeometry`).
+
+        Parameters
+        ----------
+        skydir : `~astropy.coordinates.SkyCoord`
+            Sky coordinate of the WCS reference point
+        shape : `tuple`
+            Shape of the image (Numpy axis order: y, x)
+        fov_deg : `float`
```
cdeil: I would suggest to change this option to `fov` and then in the code do this:
```
from astropy.coordinates import Angle
fov = Angle(fov)
```
This will force the user to be explicit about units of inputs (lead to more readable user scripts) and allow e.g. to pass strings like "3 deg" (i.e. users don't have to instantiate an Angle object before calling.

hips/utils/testing.py
```diff
@@ -58,5 +57,17 @@ def make_test_wcs_geometry(case=0):
             shape=(1000, 2000), coordsys='GAL',
             projection='AIT', cdelt=0.01, crpix=(1000, 500),
         )
+    elif case == 3:
```
cdeil: I suggest you move these cases simply to unit tests of `WCSGeometry` in `TestWCSGeometry`.
I.e. have test methods that instantiate a WCSGeometry by calling this method and then assert some properties of the resulting object.

The only reason for this `make_test_wcs_geometry` was to have something available for other tests.
We should keep the number of cases here as small as possible, i.e. simple unit tests for `WCSGeometry` go to `TestWCSGeometry`.

hips/utils/wcs.py
```diff
@@ -124,6 +124,30 @@ def create(cls, skydir: SkyCoord, shape: tuple, coordsys: str = 'CEL',
...
+            Field of view in degrees
+        coordsys : `str`
+            Coordinate system
+        projection : `str`
+            Projection of the WCS object
```
cdeil: Add an `Examples` section with an example (maybe the same one as you have for `create`?
I would also suggest that you use this more convenient method here:
https://hips.readthedocs.io/en/latest/getting_started.html#make-a-sky-image

hips/utils/wcs.py
adl1995: Do you mean to pass 'icrs' and 'galactic' instead of 'CEL' and 'GAL'?

cdeil: Yes.

I would suggest the end-user API of the hips package always has "icrs" and "galactic", and where needed we convert to other strings under the hood (e.g. to compare with `hips_frame`).

adl1995:
@cdeil - I just made another commit. Hopefully this resolves all the issues.

cdeil:
See https://travis-ci.org/hipspy/hips/jobs/249595428#L1148
(always run tests locally before pushing)

---

Before we merge I have two questions on user interface, @tboch and @adl1995 - please comment:
- The names `create` and `create_simple` aren't great. I don't have a suggestion for better names. Thoughts?
- Maybe we should take separate arguments `width` and `height` or `nx` and `ny` for the number of pixels? (In https://github.com/hipspy/hips/issues/39#issuecomment-311889775 @tboch commented that he's taking width and height in Aladin Lite). I think if we take a `shape`, especially with the current order `(ny, nx)`, but even if we flip, users will get it wrong all the time . So here I'm in favour of either `width` and `height` or `nx` and `ny` instead of `shape`.

hips/utils/wcs.py
```diff
@@ -124,6 +126,54 @@ def create(cls, skydir: SkyCoord, shape: tuple, coordsys: str = 'CEL',
...
+                      coordsys: str = 'icrs', projection: str = 'AIT') -> 'WCSGeometry':
+        """Create WCS object programmatically using field of view (`WCSGeometry`).
+
+        Parameters
+        ----------
+        skydir : `~astropy.coordinates.SkyCoord`
+            Sky coordinate of the WCS reference point
+        shape : `tuple`
+            Shape of the image (Numpy axis order: y, x)
+        fov: `str`
```
cdeil: `fov` can be `Angle` or `str`, not just `str`.

adl1995:
@cdeil - How about `create_from_fov`? I'm fine with taking two separate arguments instead of `shape`.

hips/utils/wcs.py
cdeil: @adl1995 - Please address this one remaining inline comment. Otherwise this is 👍  from my side.

@tboch - Assiging to you for final review.

adl1995:
I made another commit. However, as I pulled the latest code from `master` branch, I now have a commit named:

> Merge branch 'master' of https://github.com/hipspy/hips into create_wcs

cdeil:
The common way to avoid these merge commits is to "rebase" instead of "merging".
Please read the Astropy dev docs (or other resources) explaining this and do that in the future.
For this PR: either leave as-is, or fix it up (e.g. by squashing commits, or doing a hard reset on the commit before the merge commit).

adl1995:
@cdeil - I have removed the merge commit.

tboch:
Things look fine for me.
I like @cdeil 's suggestion to split the `shape` argument into `width` and `height`. Do we leave that for another pull request?

cdeil:
As discussed in telcon: suggest to keep `Shape` attribute on `WCSGeometry` as-is, but only to take parameters `nx` and `ny` separately in constructors and make and store the `Shape` on self.

cdeil:
Correction: call it width and height (both in the constructor arguments, and in the `Shape` attributes.

adl1995:
@cdeil I have made a commit updating the `shape` parameter.

docs/getting_started.rst
```diff
@@ -20,10 +20,10 @@ To make a sky image with the `hips` package, follow the following three steps:
     from astropy.coordinates import SkyCoord
     from hips.utils import WCSGeometry
 
-    geometry = WCSGeometry.create(
+    geometry = WCSGeometry.create_simple(
          skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
-         shape=(1000, 2000), coordsys='GAL',
-         projection='AIT', cdelt=0.01, crpix=(1000, 500),
+         shape=(1000, 2000), fov="3 deg",
```
cdeil: This still calls with `shape`, needs update, no?

hips/utils/testing.py
```diff
@@ -43,19 +42,19 @@ def make_test_wcs_geometry(case=0):
     if case == 0:
         return WCSGeometry.create(
             skydir=SkyCoord(3, 4, unit='deg', frame='galactic'),
-            shape=(2, 3), coordsys='GAL',
+            height=2, width=3, coordsys='galactic',
```
cdeil: I'd suggest we always write in the order `width`, then `height`, even if it doesn't matter when calling with kwargs.  Please change here and below.

hips/utils/wcs.py
```diff
@@ -46,20 +52,20 @@ class WCSGeometry:
...
     NAXIS : 0  0
     >>> wcs_geometry.shape
-    Shape(ny=10, nx=20)
+    Shape(height=10, width=20)
     """
     WCS_ORIGIN_DEFAULT = 0
     """Default WCS transform origin, to be used in all WCS pix <-> world calls."""
 
-    def __init__(self, wcs: WCS, shape: tuple) -> None:
+    def __init__(self, wcs: WCS, height: int, width: int) -> None:
```
cdeil: I'd suggest we always use the order: width, then height , i.e. please change this here.

hips/utils/wcs.py
```diff
@@ -78,35 +84,38 @@ def pix_to_sky(self, x, y) -> SkyCoord:
...
         ----------
         skydir : `~astropy.coordinates.SkyCoord`
             Sky coordinate of the WCS reference point
-        shape : `tuple`
-            Shape of the image (Numpy axis order: y, x)
-        coordsys : `str`
+        height : `int`
+            Height of the image
+        width : `int`
+            Width of the image
```
cdeil: Mention "in pixels" or "number of pixel" to make it perfectly clear that this is not width e.g. as an angle on the sky.

hips/utils/tests/test_wcs.py
```diff
@@ -1,18 +1,35 @@
...
+            skydir=SkyCoord(0, 0, unit='deg', frame='icrs'),
+            height=1000, width=2000, fov="3 deg",
+            coordsys='icrs', projection='AIT'
+        )
+        c = wcs_geometry.center_skycoord
+
+        assert c.frame.name == 'icrs'
+        assert_allclose(c.ra.deg, 359.24, atol=1e-2)
+        assert_allclose(c.dec.deg, -0.74, atol=1e-2)
+        assert_allclose(self.wcs_geometry.wcs.wcs.crpix, [500.,  1000.])
```
cdeil: Is this correct? Doesn't `crpix` have order `x`, `y` and should be `1000`, `500` here?

@adl1995 - You could check by making a test image locally and e.g. filling the data with zeros or with coordinate values like we do somewhere else, and then writing to FITS and opening it up.

hips/utils/wcs.py
```diff
@@ -4,16 +4,20 @@
...
 from astropy.wcs.utils import pixel_to_skycoord, wcs_to_celestial_frame
 
 __all__ = [
     'WCSGeometry',
 ]
 
-__doctest_skip__ = ['WCSGeometry']
+__doctest_skip__ = [
+    'WCSGeometry',
+    'WCSGeometry.*'
```
cdeil: Why is this second entry `WCSGeometry.*` needed here?
Shoudn't the first `WCSGeometry` entry lead to all doctests in this class being skipped?
Please consult the Astropy docs on this point if you're not sure.

adl1995: This is required to skip doctests for all the methods within `WCSGeometry`, in this case the `create_simple` `@classmethod`, more [here](http://docs.astropy.org/en/stable/development/testguide.html#skipping-doctests).

cdeil: I find that weird that listing a class doesn't mean skip all doctests (including methods) in that class. But OK, if that's how it works, then keep as is of course.

adl1995: One alternative would be to use `*` in place of both `WCSGeometry` and `WCSGeometry.*` which would skip all the tests. Do you think that is more suitable?

cdeil: Well generally, we should probably first try to write docs examples where the docstests pass.
I think you already saw how if some docs example is untested, it will usually break quickly as we change and refactor the code.
So for now, we have to manually go through and review the docs examples from time to time, e.g. before a release.

For now, do as you like here: fix up the docs example so that it works, or skip in whatever way you like (maybe best to make it as narrow as possible, i.e. just list the method with the doctest you can't get to work)?

hips/utils/tests/test_wcs.py
adl1995: @cdeil How exactly can I check this? I created an image using the survey in `test_simple.py` and loaded the FITS image in Aladin Desktop. It looks okay to me, but is there anything else that I should be checking?

cdeil: E.g. ds9, but I think also Aladin will show the mouse position in pixel and world coordinates when you hover over the image.

If you go to the reference point of the projection (the `crval` sky coordinates), what pixel coordinates do you see?
It should be `x=1000` and `y=500`, but I suspect you'll find it switched.

cdeil: And you can also try to draw a sky image with this geometry like in this high-level docs example we have. Does it come out correctly or are there issues?

adl1995: I just tried this. I created the `WCSGeometry` object this way:

`geometry = WCSGeometry.create(skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),      width=2000, height=1000, coordsys='galactic', projection='AIT', cdelt=0.01, crpix=(1000, 500))`

i.e. reversed the order `crpix`. This is the result I'm getting in Aladin:

screenshot - 060717 - 13 36 06

This should be reversed, no?

adl1995: I just tried reversing `crpix` values i.e.

geometry = WCSGeometry.create(skydir=SkyCoord(0, 0, unit='deg', frame='galactic'), width=2000, height=1000, coordsys='galactic', projection='AIT', cdelt=0.01, crpix=(1000, 500))

But I still get the same result as above.

cdeil: The test and my comment were about `create_simple` and how it fills `crpix`.
Please use that for checking.

It's hard for me to comment here without seeing which code exactly you ran.
Maybe you could make a notebook that has code like the current getting started, but using `create_simple` and at the end plot the sky image and a marker using `crpix` and `crval` to show where it is.
Then I could tell you if there currently is a bug or not (or probalby you'll understand the situation yourself when doing this).

adl1995:
@cdeil I have updated the test cases. The build is passing now. However, I did make some changes on my local notebook, but the results looked identical to me. Please verify before merging.

cdeil:
@adl1995 - Thanks for fixing the bug where the values in `crpix` were switched in `create_simple`.

I've attached two commits here, one with minor cleanup, and one fixing this test fail:
https://gist.github.com/cdeil/04226e02ea9c924d24ae91d53e75dcf5

For now I've simply updated the reference value, I'm not sure which is correct.
Also, the high-level docs example is broken now (stuck fetching tiles forever).

I think we should just merge the open PRs in today, especially the tile order computation step, and then re-start debugging.

Issue #43, opened 2017-06-30
Better algorithm to get list of HiPS tiles


cdeil:
With the one example for `make_sky_image` that we have in the docs we get this iamge:
https://github.com/hipspy/hips/pull/30#issuecomment-312290956

There's two tiles missing at the edge, because the algorithm we have is to do a disk query with center at the center of the image, and radius such that all pixels in the sky image are covered by the disk, and then it can happen that tile centers are outside that disk:
http://hips.readthedocs.io/en/latest/api/hips.utils.compute_healpix_pixel_indices.html

The current algorithm is following the description of:
http://hips.readthedocs.io/en/latest/drawing_algo.html#naive-algorithm

@tboch - Do you have a suggestion for a better algorithm?

tboch:
Right, it seems that query_disc method sometimes misses some pixels, especially for small orders (large field of view). In Aladin Lite, the workaround I used was to extend the radius by a factor 1.5:
`radius *= 1.5`
This will return additional unneeded tiles, but they can be filtered out easily using method described in step 5 of http://hips.readthedocs.io/en/latest/drawing_algo.html#naive-algorithm

cdeil:
@tboch - I was thinking about this a bit: extending the radius by 1.5 can still result in missing tiles if the sky image is small, no?

Here's an alternative algorithm: change `compute_healpix_pixel_indices` and call
http://healpy.readthedocs.io/en/latest/generated/healpy.pixelfunc.ang2pix.html
for the coordinates of all the pixels in the image (we already have those as `pixel_coords = wcs_geometry.pixel_skycoords` in the function currently), followed by making making a unique list.

That should always give the exact list of tiles needed, and also means we don't need the tile list trimming step that's needed with the disk query. When drawing it could also be useful to have this "image" of `ipix` knowing which tile will / should be used to draw on a given pixel in the sky image.

Possible drawbacks of that method are that it could be slower (`ang2pix` and computing `unique`) or memory intensive (extra image array of `ipix`).

I think my suggestion would be to implement the algorithm with `ang2pix` first, because it should be just <~ 5 lines of code, and always yield correct results.

@tboch - Thoughts?

tboch:
Yes indeed, this should always return the exact correct list of needed tiles. Please @adl1995  go ahead with @cdeil 's suggestion, we'll see later how we cope if computation is too slow.

adl1995:
[...]

PR #50, opened 2017-07-03
Make improvements to compute_healpix_pixel_indices utility function

adl1995:
Made improvements to the algorithm as outlined in #43.

### DIFF
--- a/hips/utils/healpix.py
+++ b/hips/utils/healpix.py
@@ -69,11 +69,12 @@ def boundaries(nside: int, pix: int, nest: bool = True) -> tuple:
     return theta, phi
 
 
-def compute_healpix_pixel_indices(wcs_geometry: WCSGeometry, order: int, healpix_frame: str = None) -> np.ndarray:
+def compute_healpix_pixel_indices(wcs_geometry: WCSGeometry, order: int, healpix_frame: str) -> np.ndarray:
     """Compute HEALPix pixels within a minimal disk covering a given WCSGeometry.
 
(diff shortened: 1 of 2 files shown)


cdeil:
The results look good to me.

You have to update the function docstring. It currently describes the old algorithm (very briefly, with one sentence summary).

Concerning implementation, I have to admit that I find the current `np.bincount` followed by `np.nonzero` very hard to understand (even after looking at https://docs.scipy.org/doc/numpy/reference/generated/numpy.bincount.html).
@adl1995 - Could you please try https://docs.scipy.org/doc/numpy/reference/generated/numpy.unique.html as suggested here? https://stackoverflow.com/a/25943480/498873 It should give identical results, be one line shorter/simpler and my guess is that it's also faster (although we shouldn't spend time on benchmarking that now, we'll do that once we have a complete package to find the bottlenecks: probably it's going to be the projective transform call in drawing)

@tboch - How do you want to handle updates to https://hips.readthedocs.io/en/latest/drawing_algo.html#naive-algorithm ?
Would it be OK if @adl1995 updates that description to reflect the algorithm we're implementing? (e.g. that page currently still says that tiles should be split in triangles, and that query_disk is called)

adl1995:
@cdeil I have made the changes. Please check.

tboch:
@cdeil  - I'm fine with @adl1995 updating https://hips.readthedocs.io/en/latest/drawing_algo.html#naive-algorithm .

cdeil:
OK, I'm merging this now so that you can more easily continue with your other PRs.

@tboch - If you ever find that I merged something too quickly, please just leave another comment or open a new issue.

Issue #51, opened 2017-07-03
Method HipsSurveyProperties.tile_access_url returns incorrect results


tboch:
Current method HipsSurveyProperties.tile_access_url is:
```python
    @property
    def tile_access_url(self):
        """Tile access URL for a HiPS surveys"""
        return self.base_url + '/Norder' + str(self.hips_order) + '/Dir0/'
```

I see 2 problems with this:

- order is always the maximum order of the HiPS
- the directory number is hard-coded to **Dir0** which will lead to wrong results once we try to access tiles with order>4

Thus, I'd suggest adding 2 arguments and adjust the code so that the method signature becomes:
```python
    def tile_access_url(self, order: int, tile_index: int):
    ...
```

tboch:
Fixed by @adl1995 in [#52](https://github.com/hipspy/hips/pull/52)

PR #52, opened 2017-07-03
Update tile_access_url method

adl1995:
Updated the `tile_access_url` method as mentioned in #51.

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -97,7 +97,7 @@ def fetch_tiles(healpix_pixel_indices: np.ndarray, order: int, hips_survey: Hips
             frame=hips_survey.astropy_frame,
             file_format='fits',
         )
-        tile = HipsTile.fetch(tile_meta, hips_survey.tile_access_url + tile_meta.filename)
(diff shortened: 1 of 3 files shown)


tboch:
@adl1995 - please modify your `test_tile_access_url `function to assert the URL of arguments `order=9, tile_index=54321`
As you know, tiles are grouped in directories by 10,000 items, thus these arguments should return URL `http://alasky.u-strasbg.fr/DSS/DSSColor/Norder9/Dir50000/`
This test should bring up an issue with your current code.

adl1995:
I have updated the code to compute directory number using: `np.around(ipix, decimals=-(len(str(ipix)) - 1))`. This works for `ipix = 54321`, which returns `50000`. But, for `ipix = 448`, it returns `400`, when instead, it should return `0`.

tboch:
@adl1995 - instead of `np.around(ipix, decimals=-(len(str(ipix)) - 1)) `, I suggest using `(ipix // 10000) * 10000`

adl1995:
@tboch Thanks! That worked.

PR #55, opened 2017-07-05
Add test for PNG tiles

adl1995:
This PR adds a test case for PNG tiles. The same code for jpg tiles is used for png.

### DIFF
--- a/hips/tiles/tests/test_tile.py
+++ b/hips/tiles/tests/test_tile.py
@@ -1,62 +1,70 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+import pytest
 import numpy as np
 from astropy.tests.helper import remote_data
 from numpy.testing import assert_allclose
 from numpy.testing import assert_equal
 from ..tile import HipsTile, HipsTileMeta
-
+from ...utils.testing import get_hips_extra_file, requires_hips_extra
 
 class TestHipsTile:
-    @remote_data
-    def test_fetch_read_write_fits(self, tmpdir):
-        meta = HipsTileMeta(order=6, ipix=30889, file_format='fits')
-        url = 'http://alasky.unistra.fr/2MASS/H/Norder6/Dir30000/Npix30889.fits'
-        tile = HipsTile.fetch(meta, url)
 
-        assert tile.data.shape == (512, 512)
-        assert_equal(tile.data[510][5:7], [1, 0])
-
-        filename = str(tmpdir / 'Npix30889.fits')
-        tile.write(filename)
-        tile2 = HipsTile.read(meta, filename=filename)
-
-        assert tile == tile2
+    fetch_read_pars = [
+        dict(url='http://alasky.unistra.fr/DSS/DSS2Merged/Norder3/Dir0/Npix463.fits',
+             full_path='datasets/samples/DSS2Red/Norder3/Dir0/Npix463.fits',
+             file_name='Npix463.fits', file_format='fits', order=3, ipix=463),
+        dict(url='http://alasky.unistra.fr/DSS/DSS2Merged/Norder3/Dir0/Npix463.jpg',
+             full_path='datasets/samples/DSS2Red/Norder3/Dir0/Npix463.jpg',
+             file_name='Npix463.jpg', file_format='jpg', order=3, ipix=463),
+        dict(url='http://alasky.unistra.fr/2MASS6X/2MASS6X_H/Norder6/Dir0/Npix6112.png',
+             full_path='datasets/samples/2MASS6XH/Norder6/Dir0/Npix6112.png',
+             file_name='Npix6112.png', file_format='png', order=6, ipix=6112)
+    ]
 
     @remote_data
-    def test_fetch_read_write_jpg(self, tmpdir):
-        meta = HipsTileMeta(order=6, ipix=30889, file_format='jpg')
-        url = 'http://alasky.unistra.fr/2MASS/H/Norder6/Dir30000/Npix30889.jpg'
-        tile = HipsTile.fetch(meta, url)
+    @requires_hips_extra()
+    @pytest.mark.parametrize('pars', fetch_read_pars)
+    def test_fetch_read(self, pars):
+        meta = HipsTileMeta(order=pars['order'], ipix=pars['ipix'], file_format=pars['file_format'])
+        tile_fetched = HipsTile.fetch(meta, url=pars['url'])
 
-        assert tile.data.shape == (512, 512, 3)
-        assert_equal(tile.data[510][5:7], [[0, 0, 0], [1, 1, 1]])
+        full_path = get_hips_extra_file(pars['full_path'])
+        tile_local = HipsTile.read(meta, full_path=full_path)
 
-        filename = str(tmpdir / 'Npix30889.jpg')
+        assert tile_fetched == tile_local
+
+    read_write_pars = [
+        dict(full_path='datasets/samples/DSS2Red/Norder3/Dir0/Npix463.fits', file_name='Npix463.fits',
+             file_format='fits', order=3, ipix=463, shape=(512, 512), tile_data=[3047], index=[[510], [5]]),
+        dict(full_path='datasets/samples/DSS2Red/Norder3/Dir0/Npix463.jpg', file_name='Npix463.jpg',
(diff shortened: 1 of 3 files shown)


cdeil:
Nice to see that PNG I/O works out of the box.

I left one inline comment. Otherwise, as discussed on Slack, I would suggest to refactor the tile tests to use tiles from hips-extra as much as possible, and to only leave one test fetching tiles and asserting that they are identical to tiles read locally from hips-extra, so that we rely on `@remote_data` as little as possible.
Note that currently remote_data tests don't run on travis-ci, so this might actually bring up new issues we hadn't noticed before (probably not though).

hips/tiles/tile.py
```diff
@@ -189,15 +189,13 @@ def _from_raw_data(cls, meta: HipsTileMeta, raw_data: BytesIO) -> 'HipsTile':
...
+        elif meta.file_format in {'jpg', 'png'}:
             with Image.open(raw_data) as image:
                 data = np.array(image)
             return cls(meta, data)
-        elif meta.file_format == 'png':
-            raise NotImplementedError()
         else:
             raise ValueError(f'Tile file format not supported: {meta.file_format}. '
-                             'Supported formats: fits, jpg, png')
+                              'Supported formats: fits, jpg, png')
```
cdeil: I think there's a bug here (that I introduced earlier, not you here): `meta.file_format` doesn't exist, you'd have to use either `self.meta.file_format` or better the local variable `file_format` defined above.

This is pretty common that error handling code (like in this case an invalid `file_format` isn't tested, and then the error handling code has bugs and users get super confused if they hit that code and get a chained exception.
So bottom line: even if it's annoying / quite some work, I would suggest we try to get 100% or almost 100% test coverage and add tests exercising such lines as this one.

The way to do it is to prepare a test case that will reach this line of code and then to call this method inside a `pytest.raises(ValueError)` block, which makes sure the expected error gets raises (and not a NameError like with the current buggy code):
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

hips/tiles/tests/test_tile.py
```diff
@@ -4,45 +4,103 @@
...
-        assert_equal(tile.data[510][5:7], [[0, 0, 0], [1, 1, 1]])
+        filename = str(tmpdir / 'Npix463.jpg')
+        tile.write(filename)
+        tile2 = HipsTile.read(meta, full_path=filename)
+
+        self.assert_jpg(tile, tile2)
 
-        filename = str(tmpdir / 'Npix30889.jpg')
+    @remote_data
+    def test_fetch_read_write_png(self, tmpdir):
```
cdeil: Put a better name here like e.g. `test_fetch_png`, especially remove the "write" part. Really this method is now only about testing fetch.

hips/tiles/tests/test_tile.py
```diff
@@ -4,45 +4,103 @@
 from numpy.testing import assert_allclose
 from numpy.testing import assert_equal
 from ..tile import HipsTile, HipsTileMeta
-
+from ...utils.testing import get_hips_extra_file, requires_hips_extra
 
 class TestHipsTile:
+    def assert_fits(self, tile1, tile2):
```
cdeil: I don't think it makes sense to split out these `assert_*` methods.
You call each of them twice: once from the read/write test, and once from the fetch test.

You can just copy these asserts at the end of the read / write tests, and then by simply having `tile_read == tile_fetched` at the end of the fetch tests, you'll have tested exactly the same thing, without the extra complexity of these three separate methods.

hips/tiles/tests/test_tile.py
```diff
@@ -4,45 +4,103 @@
...
-        # the data is different (for unknown reasons).
-        # TODO: Figure out what's wrong here and fix!
-        # print(tile.data.sum())
-        # print(tile2.data.sum())
-        # print((tile == tile2).all())
-        # assert tile == tile2
+        self.assert_png(tile, tile2)
+
+    @requires_hips_extra()
+    def test_read_write_fits(self, tmpdir):
```
cdeil: It would be possible to use `pytest.mark.parametrize` to cover the three cases of PNG, JPEG and FITS with just one common function. @adl1995 - If you think it's a good idea and know how to do it, you can make this change.
But the gain is very small, and there might be other asserts that we want to do specifically for each type, so I'm also OK to leave as-is.

adl1995:
@cdeil I just made another commit.

hips/tiles/tests/test_tile.py
```diff
@@ -1,46 +1,87 @@
...
         # TODO: Figure out what's wrong here and fix!
-        # print(tile.data.sum())
+        # print(tile1.data.sum())
         # print(tile2.data.sum())
-        # print((tile == tile2).all())
+        # print((tile1 == tile2).all())
+        # assert tile1 == tile2
+
+    @remote_data
+    def test_fetch_png(self, tmpdir):
```
cdeil: Suggest to change all fetch tests like this:

* Use a tile from hips-extra
* Fetch it from the raw URL like e.g. https://github.com/hipspy/hips-extra/raw/master/datasets/samples/2MASS6XH/Norder6/Dir0/Npix1517.png
* Also read it via Tile.read
* Assert tile1 == tile2
* Don't do any extra asserts or tile writing (instead do the assertions on shape or pixel values in the other, non-fetch tests below)

hips/tiles/tests/test_tile.py
```diff
@@ -1,46 +1,87 @@
...
+        meta = HipsTileMeta(order=pars['order'], ipix=pars['ipix'], file_format=pars['file_format'])
+        tile = HipsTile.read(meta, full_path)
+
+        assert tile.data.shape == pars['shape']
+        assert_equal(tile.data[510][5:7], pars['tile_data'])
+
+        filename = str(tmpdir / pars['file_name'])
+        tile.write(filename)
+        tile2 = HipsTile.read(meta, full_path=filename)
+
```
cdeil: Suggest to leave a TODO comment here, and fix the JPEG to round-trip in a later PR (like we discussed, by storing on `raw_data`)

hips/tiles/tests/test_tile.py
```diff
@@ -1,46 +1,87 @@
...
+        assert tile == tile2
+
+    read_write_pars = [
+        dict(full_path='datasets/samples/DSS2Red/Norder3/Dir0/Npix463.fits', file_name='Npix463.fits',
+             file_format='fits', order=3, ipix=463, shape=(512, 512), tile_data=[3047, 3257]),
+        dict(full_path='datasets/samples/DSS2Red/Norder3/Dir0/Npix463.jpg', file_name='Npix463.jpg',
+             file_format='jpg', order=3, ipix=463, shape=(512, 512, 3), tile_data=[[10, 10, 10], [11, 11, 11]]),
+        dict(full_path='datasets/samples/2MASS6XH/Norder6/Dir0/Npix6112.png', file_name='Npix6112.png',
+             file_format='png', order=6, ipix=6112, shape=(512, 512, 4), tile_data=[[0, 0, 0, 0], [0, 0, 0, 0]]),
```
cdeil: There should be a test for a pixel in PNG that's not all zeros (to show that it really works).
You could just change the pixel data tests to be for one pixel, and store it's coordinates also in the dicts here.

hips/tiles/tile.py
```diff
@@ -189,25 +189,23 @@ def _from_raw_data(cls, meta: HipsTileMeta, raw_data: BytesIO) -> 'HipsTile':
...
             with Image.open(raw_data) as image:
                 data = np.array(image)
             return cls(meta, data)
-        elif meta.file_format == 'png':
-            raise NotImplementedError()
         else:
-            raise ValueError(f'Tile file format not supported: {meta.file_format}. '
-                             'Supported formats: fits, jpg, png')
+            raise ValueError(f'Tile file format not supported: {self.meta.file_format}. '
+                              'Supported formats: fits, jpg, png')  # pragma: no cover
```
cdeil: Suggest to add a separate test actually triggering this ValueError and asserting that it comes out via `pytest.raises`, i.e. don't put `#pragma: no cover`

adl1995:
@cdeil I made a commit fixing the issues discussed earlier.

cdeil:
Thanks!

Issue #54, opened 2017-07-04
Get tile dimension from properties file


tboch:
In current version, tiles are assumed to be 512x512 pixels.This is not always the case ; for instance, the HiPS described by http://alasky.unistra.fr/PLANCK/HFI143/properties has some 256x256 pixels tiles. Here is the kind of cutout I get for this HiPS:
<img width="524" alt="capture d ecran 2017-07-04 a 09 26 41" src="https://user-images.githubusercontent.com/5598168/27819802-28b047b8-609c-11e7-9d5c-f133e9d2883e.png">

Tile dimension should be read from properties key `hips_tile_width` if it exists, and otherwise defaults to 512.

cdeil:
I think it would be good to have tests that show that our package works for this case.
@tboch - How about adding that as an example dataset to `hips-extra`?

adl1995:
What is the best way to handle this? By instantiating a `HipsSurveyProperties` object inside `HipsTile` and reading `hips_tile_width` from there? Or should the user just pass it in after reading it from the appropriate location?

tboch:
@cdeil - you are right, I have added a new sample `Planck-HFI143` to `hips-extra`
It has 256x256 pixels tiles

tboch:
@adl1995 - thinking again about this, I realize we can do without the properties file to get the tile dimension. Once the tile has been fetched, we just have to look at the data shape.

@cdeil - thoughts about this?

cdeil:
Isn't the tile width needed already to compute the tile list at a point in time where no tiles have been fetched yet? At that point at least it would have to be accessed from `HipsSurveyProperties`, no?

I don't have a clear picture yet when we should take meta info from the `HipsTile` or `HipsTileMeta` or from `HipsSurveyProperties`. My suggestion would be to just plow on and make all the major use-cases work, and then to do a session together (or really any time) reviewing the code we have, trying to re-organise things to be as simple and clear as possible.

tboch:
Well, you are right as the tile width will change the resolution computation (I hadn't thought about this first), thus the order of the tiles to download.
But for now, we can just ignore this, and resolve tile width when reading the tiles data array.

cdeil:
We also have `tile_width` as an attribute on [HipsTileMeta](https://hips.readthedocs.io/en/latest/api/hips.HipsTileMeta.html#hips.HipsTileMeta).
It's not clear to me if that is a good idea or not.
[...]

PR #56, opened 2017-07-05
Remove tile_width from HipsTileMeta

adl1995:
This PR addresses #54.

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -14,7 +14,7 @@ __all__ = [
 
 
 # TODO: Fix type annotation issue
-def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any]) -> np.ndarray:
+def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any], hips_survey: HipsSurveyProperties) -> np.ndarray:
     """Draw sky image using the simple and quick method.
 
     Parameters
@@ -23,6 +23,8 @@ def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any]) 
         An object of WCSGeometry
     tiles : List[HipsTile]
         A list of HipsTile
+    hips_survey : `~hips.HipsSurveyProperties`
+        HiPS survey properties
 
     Returns
     -------
@@ -31,7 +33,7 @@ def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any]) 
     """
     image = np.zeros(geometry.shape)
     for tile in tiles:
-        painter = SimpleTilePainter(geometry, tile)
+        painter = SimpleTilePainter(geometry, hips_survey, tile)
         image += painter.warp_image()
     return image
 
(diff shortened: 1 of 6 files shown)


hips/draw/simple.py
```diff
@@ -32,7 +34,7 @@ def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any])
     image = np.zeros(geometry.shape)
     for tile in tiles:
         painter = SimpleTilePainter(geometry, tile)
-        image += painter.warp_image()
+        image += painter.warp_image(hips_survey.tile_width)
```
cdeil: Suggest to pass the `hips_survey` object to `SimpleTilePainter.__init__`, store it there as an attribute on self, and then access it where needed.

cdeil:
Looks like you forgot to update one test to the API change here:
https://travis-ci.org/hipspy/hips/jobs/250275541#L1144

hips/tiles/tile.py
```diff
@@ -83,21 +79,19 @@ def nside(self) -> int:
...
-             [0, 0]],
-        )
-
-    @property
     def skycoord_corners(self) -> SkyCoord:
         """Corner values for a HiPS tile"""
         theta, phi = boundaries(self.nside, self.ipix)
         return SkyCoord(phi, np.pi / 2 - theta, unit='radian', frame=self.frame)
 
+    def dst(self, tile_width) -> np.ndarray:
```
cdeil: @tboch  and I suggest to move this to a property on the `SimpleHipsTilePainter` class (or a standalone function in the `draw.py` code).

adl1995:
@cdeil Please check the latest commit. It addresses the issues discussed during the telecon today.

cdeil:
Rebase, please.

adl1995:
@cdeil just rebased.

cdeil:
Did you force push after rebasing? 

The last commit I see here is dd6e3d9e50e290c3d9c1fe602a10bd5f5f5b9f42 from yesterday.

adl1995:
Yes. I also squashed a commit. I see there is a failing Travis CI build. I will update and push again.

adl1995:
@cdeil Just updated. Please check.

hips/draw/simple.py
```diff
@@ -45,20 +47,32 @@ class SimpleTilePainter:
...
+    def __init__(self, geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile: HipsTile) -> None:
         self.geometry = geometry
+        self.hips_survey = hips_survey
         self.tile = tile
 
     @property
+    def dst(self) -> np.ndarray:
+        """Destination array for projective transform"""
+        return np.array(
+            [[self.hips_survey.tile_width - 1, 0],
```
cdeil: I'd suggest to create a local variable
```
width = self.hips_survey.tile_width
```
and then use it instead of having that lengthy expression 4 times.

hips/tiles/surveys.py
```diff
@@ -128,7 +128,13 @@ def base_url(self) -> str:
         """HiPS access url"""
         return self.data['moc_access_url'].rsplit('/', 1)[0]
 
+    @property
+    def tile_width(self) -> int:
+        """HiPS tile width"""
+        return int(self.data['hips_tile_width']) or 512
+
     def directory(self, ipix: int) -> int:
+        """Directory containing HiPS tile(s)"""
```
cdeil: This description isn't great. This returns an int, not a directory.
Can you be more specific?

cdeil:
This is almost ready to go.

I left two minor inline comments, and looking at #54 to fully address it I think it would be nice if you add a test where you read the properties file and assert that the tile width is 256 for this case:
https://github.com/hipspy/hips/issues/54#issuecomment-312840189
(and later on also use that for drawing tests)

cdeil:
Thanks!

Issue #47, opened 2017-07-03
Compute HiPS order for simple drawing method


cdeil:
I noticed that we're still completely missing step 1 of
https://hips.readthedocs.io/en/latest/drawing_algo.html
i.e. to compute the HiPS oder for the simple drawing method.

At the moment we're simply taking `hips_survey.hips_order` in `make_sky_image` (see [here](https://hips.readthedocs.io/en/latest/_modules/hips/draw/simple.html#make_sky_image), which (I think) will take the `hips_order` from the HiPS survey properties file, and will be the maximum resolution order available. Usually this is not the order one wants to use for drawing, i.e. for many cases like the large sky iamge we have this would fetch thousands or millions of tiles.

There's a few different ways to compute a pixel scale for the sky image (see e.g. [here](https://gammapy.readthedocs.io/en/latest/api/gammapy.image.SkyImage.html#gammapy.image.SkyImage.wcs_pixel_scale)) and then there's the question how exactly to transform that to a HEALPix order.

@tboch - Can you give specifics how exactly the order to use for drawing is computed in Aladin Lite? I think maybe we should just use the same formula here, at least for now as starting point.

tboch:
I suggest the following function:
```python

import numpy as np
from astropy.wcs.utils import proj_plane_pixel_scales

def get_order(geometry, hips_survey):
    """ return the tile order suited for the given geometry and hips_survey    """

    # resolution in deg/pixel
    geom_resolution = np.min(proj_plane_pixel_scales(geometry.wcs))
    # we want to find the smaller tile order with a resolution equal or better than geom_resolution
    tile_size = 512 # this must be adjusted according to value of property hips_tile_width (if it exists)
    tile_order = np.log2(tile_size)
    full_sphere_area = 4 * np.pi  * np.square(180 / np.pi)
    for candidate_tile_order in range(3, 29 + 1): # 29 is the max order supported by healpy
        tile_resolution = np.sqrt(full_sphere_area / 12 /  4**(candidate_tile_order + tile_order))
        
        if (tile_resolution<=geom_resolution):
            break

    # order can't be larger than hips_order
    return np.min([candidate_tile_order, hips_survey.hips_order])
```

adl1995:
@tboch `candidate_tile_order` is not defined in the `return` statement: `np.min([candidate_tile_order, hips_survey.hips_order])`, what should I set its value to? Should it be an array in the range (3, 29 + 1)?

tboch:
[...]

PR #58, opened 2017-07-06
Add function for computing HiPS order

adl1995:
This PR addresses issue #47.

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -2,6 +2,7 @@
 """HiPS tile drawing -- simple method."""
 from typing import Generator, Any
 import numpy as np
+from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
 from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta
 from ..utils import WCSGeometry, compute_healpix_pixel_indices
@@ -116,6 +117,22 @@ def fetch_tiles(healpix_pixel_indices: np.ndarray, order: int, hips_survey: Hips
         yield tile
 
 
+def get_order(geometry: WCSGeometry, hips_survey: HipsSurveyProperties) -> int:
+    """Compute the tile order suited for the given geometry and hips_survey"""
+
+    resolution = np.min(proj_plane_pixel_scales(geometry.wcs))
+
+    tile_order = np.log2(hips_survey.tile_width)
+    full_sphere_area = 4 * np.pi * np.square(180 / np.pi)
+    for candidate_tile_order in range(3, 29 + 1):
+        tile_resolution = np.sqrt(full_sphere_area / 12 / 4**(candidate_tile_order + tile_order))
+
+        if tile_resolution <= resolution:
+            break
+
+    return np.min([candidate_tile_order, hips_survey.hips_order])
+
+
 def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties) -> np.ndarray:
     """Make sky image: fetch tiles and draw.
 
@@ -133,13 +150,14 @@ def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties) -> 
     data : `~numpy.ndarray`
         Output image pixels
     """
+    order = get_order(geometry, hips_survey)
     healpix_pixel_indices = compute_healpix_pixel_indices(
         wcs_geometry=geometry,
-        order=hips_survey.hips_order,
+        order=order,
         healpix_frame=hips_survey.astropy_frame,
     )
     # TODO: this isn't a good API. Will become better when we have a cache.
-    tiles = fetch_tiles(healpix_pixel_indices, hips_survey.hips_order, hips_survey)
+    tiles = fetch_tiles(healpix_pixel_indices, order, hips_survey)
 
     image_data = draw_sky_image(geometry, tiles, hips_survey)
 
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -10,9 +10,13 @@ from ..utils import WCSGeometry, compute_healpix_pixel_indices
 __all__ = [
     'draw_sky_image',
     'make_sky_image',
-    'SimpleTilePainter'
+    'SimpleTilePainter',
+    'compute_matching_hips_order'
 ]
 
+__doctest_skip__ = [
+    'compute_matching_hips_order',
+]
 
 # TODO: Fix type annotation issue
 def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any], hips_survey: HipsSurveyProperties) -> np.ndarray:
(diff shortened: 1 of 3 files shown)


tboch:
@adl1995 it would be better if you could add a test asserting a few typical cases for this function.

adl1995:
@tboch I just added the test case.

cdeil:
@adl1995 - This test should not use `@remote_data`. OK to load something from `hips-extra` if you want.

Also -- this function should have at least 3 test cases (probably use `pytest.mark.parametrize`, for the cases where the min and max order of tiles availalbe is used, and something in between.

You can simply change the geometry fov (zoomin in and out) to trigger the three cases.

cdeil:
I wanted to understand what this function does just now, because it's a core piece of computation in our package.

This is as far as got:
```
def get_hips_order(geometry: WCSGeometry, hips_survey: HipsSurveyProperties) -> int:
    """Compute the tile order suited for the given geometry and hips_survey"""
    # Sky image angular resolution (pixel size in deg)
    resolution = np.min(proj_plane_pixel_scales(geometry.wcs))
    # Compute HiPS order so that tile pixel resolution will be as good as sky image resolution
    desired_order = _get_hips_order_for_resolution(hips_survey.tile_width, resolution)
    # Return the desired order, or the highest resolution available
    # Note that HiPS never has resolution less than 3,
    # and that limit is handled in _get_hips_order_for_resolution
    return np.min([desired_order, hips_survey.hips_order])


def _get_hips_order_for_resolution(tile_width, resolution):
    """TODO document a bit what this function does."""
    tile_order = np.log2(tile_width)
    full_sphere_area = 4 * np.pi * np.square(180 / np.pi)
    for candidate_tile_order in range(3, 29 + 1):
        tile_resolution = np.sqrt(full_sphere_area / 12 / 4 ** (candidate_tile_order + tile_order))

        if tile_resolution <= resolution:
            break

    return candidate_tile_order
```

The code does the same as what Thomas did, but it's (I hope) a bit easier to understand what the code does in that form. My suggestion would be to put it like that in the package.

@adl1995 or @tboch - Can you write two or three lines of comments that explain how `_get_hips_order_for_resolution` works?

I think it would be good to have this well-documented, just in case there's ever an issue with the algorithm in the future, and we or others will have to revisit it.

And @adl1995 - The most important thing is to have tests that establish the current behaviour / output of that function. Please add a parametrized test for this.

adl1995:
@cdeil I have updated the function and test cases, however, I can't use `fov` until #49 gets merged.

cdeil:
@adl1995 - Next steps here:

- rebase this PR on latest master
- add the HiPS order computation as the first line in `make_sky_image`
- make the high-level docs example work again (maybe it just will if you add the order computation)
- the parametrized tests you've added here don't test the important things of the function you're adding. Instead you can use one HipS, but parametrize the `geometry` to have different pixel sizes, and then for smaller sky image pixels, a higher HiPS order should be computed by `get_order`. Is it clear what the goal is here with testing this function? If no, please ask!

cdeil:
Please also add a parametrized test (in addition to the one you have) that calls `_get_hips_order_for_resolution` directly for a few tile widths and resolutions, and asserts on the returned order.

adl1995:
@cdeil I made another commit updating the test cases.

hips/draw/tests/test_simple.py
```diff
@@ -53,5 +55,34 @@ def test_make_sky_image():
     assert_allclose(data[200, 994], 3717.10091363)
     assert_allclose(data[200, 995], 3402.55292158)
 
+get_hips_order_pars = [
```
cdeil: To me, `get_hips_order_pars` sounds like a function. This is just a list, so suggest to rename to `hips_order_pars`.

hips/draw/tests/test_simple.py
```diff
@@ -53,5 +55,34 @@ def test_make_sky_image():
...
+    dict(order=7, width=2000, height=1000, fov="3 deg"),
```
cdeil: To test the functionality here you can simply use one fixed `width` and `height` (e.g. 200 and 100). All you need to vary and list here is some parameter that changes the image pixel size like e.g. `fov`.

hips/draw/tests/test_simple.py
```diff
@@ -53,5 +55,34 @@ def test_make_sky_image():
...
+    full_path = get_hips_extra_file('datasets/samples/2MASS6XH/properties')
+    hips_survey = HipsSurveyProperties.read(filename=full_path)
+    geometry = WCSGeometry.create_simple(
+        skydir=SkyCoord(0, 0, unit='deg', frame='icrs'),
+        width=pars['width'], height=pars['height'], fov=pars['fov'],
+        coordsys='icrs', projection='AIT'
+    )
+    assert get_hips_order(geometry, hips_survey) == pars['order']
+
+get_hips_order_for_resolution_pars = [
```
cdeil: Same here: rename `get_hips_order_for_resolution_pars` -> `hips_order_for_resolution_pars` or something else that doesn't sound like a function (because it is a list).

hips/draw/tests/test_simple.py
```diff
@@ -53,5 +55,34 @@ def test_make_sky_image():
...
+    dict(tile_width=512, resolution=12.32, order=3),
+    dict(tile_width=256, resolution=160.22, order=3),
```
cdeil: You should choose cases that lead to different `orders` and test the extremes, i.e. corner / edge cases of the algorithm in that function.
At the moment you only chose values that will result in `order=3`.

hips/draw/simple.py
```diff
@@ -116,6 +117,32 @@ def fetch_tiles(healpix_pixel_indices: np.ndarray, order: int, hips_survey: Hips
         yield tile
 
 
+def get_hips_order(geometry: WCSGeometry, hips_survey: HipsSurveyProperties) -> int:
+    """Compute the tile order suited for the given geometry and hips_survey"""
```
cdeil: Add `Parameters` and `Returns` sections documenting the inputs and outputs.
Maybe even an `Examples` section with one working example so that users can see how it works in the docs?

adl1995:
@cdeil The high-level docs example is already working. I tried it locally.

cdeil:
@adl1995 - Are you sure? For me it's broken in master (hangs indefinitely trying to fetch a ton of tiles). Maybe you don't see it here and have to rebase on master to see it.

The first step that computes the HipS order to use for drawing to use is missing, so it's expected that it doesn't work for most cases, no?

My suggestion would be that you rebase here to see the issue, and then fix it in this PR.

adl1995:
@cdeil I tested it in my currently branch `get_order`, in which the first step of `make_sky_image` is computing the HiPS order (https://github.com/hipspy/hips/pull/58/files#diff-166822ec0a0c3100c6f9ffdd92c4b76eR163), so it works.

cdeil:
@adl1995 - Apologies. I missed the `order = get_hips_order(geometry, hips_survey)` line. That's perfect.

Probably this is ready to be merged then.

Do you understand why drawing tests still work without modifications?
Probably somewhat by chance, or because we don't have any drawing test that relies on the `order = get_hips_order(geometry, hips_survey)` step?
Can you add one? Do you want to do it here or in a follow-up PR?

adl1995:
@cdeil I made some further changes adding a doctest example and updating the test case. The `draw_sky_image` function does not rely on the `order` parameter, so I'm not sure about your previous comment.

cdeil:
This PR fixes a very important bug in the `make_sky_image` function, yet the test still passes unchanged: https://github.com/hipspy/hips/blob/master/hips/draw/tests/test_simple.py#L46

This clearly shows that the previous test wasn't very good and you should update it to something that fails if you comment out the `get_hips_order(geometry, hips_survey)` line, and succeeds after adding that fix in `make_sky_image`, no?

adl1995:
@cdeil I haven't been able to find a survey which does not work without the `get_hips_order` function. I tried several from the `hips-extra` repo which support FITS tiles.

cdeil:
If you go to the master branch and run the high-level docs example, and add a few print statements in https://github.com/hipspy/hips/blob/master/hips/draw/simple.py#L138 to show what `hips_survey.hips_order` is and `len(healpix_pixel_indices)`, what do you get?

I expect that you'll find that this hips_order is the maximum availabe tile resolution of the Hips and you'll get 1000s or more of tiles, i.e. draw at the wrong order.
Please try to find such a case and let me know if you don't succeed.

adl1995:
@cdeil I just tried this and the list of tiles is 392. Should I use the same example in the test case?

adl1995:
@cdeil I made some changes. Please review.

adl1995:
@cdeil Can you please merge this pull request (if no changes are required)? I want to start refactoring the `SimpleTilePainter` class.

cdeil:
OK, merging now. Thanks!

Issue #53, opened 2017-07-04
HipsSurveyProperties base_url fails for some surveys


tboch:
`base_url` should be determined using the `hips_service_url` keyword from the properties file.
`moc_access_url` currently used points to the coverage of the HiPS and this property might be missing. example: http://skies.esac.esa.int/pla/HFI_SkyMap_100_2048_R2_02_full_HiPS/properties doesn't have a `moc_access_url` property.
On the other hand, `hips_service_url` will always be found.

cdeil:
Apologies, this is something I changed / broke in https://github.com/hipspy/hips/commit/8af90581b6d63c75d939226de4d23116609312b5 .

The reason I did was that I was using http://alasky.unistra.fr/DSS/DSS2Merged/properties for the example, and that file **does not** have a `hips_service_url` key!???

tboch:
OK, let me check this with my CDS colleagues. Reading again the HiPS document, it seems that `hips_service_url key` is not required, which I don't understand for now.

cdeil:
A practical solution for now to get it to work for as many HiPS as possible might be to use `hips_service_url` and to fall back to `moc_access_url` (and leave a comment to one case where each is missing to explain why we do that). But yes, @adl1995 - Please wait for Thomas to make a decision here before making a PR.

tboch:
Alright, for historical and pragmatic reasons, things are more complex than I first thought. The algorithm we could use is:

- use `hips_service_url` if it exists
- else fall back to `moc_access_url`
- if both keys are missing and if properties URL ends with _'/properties'_ , return the URL without the _'/properties'_ suffix

adl1995:
@tboch Is the properties URL present under a key named `properties_url`? I haven't been able to find this in any of the files in `hips-extra` repository.

tboch:
@adl1995 no, there isn't such key and I'm not sure why we would need one. What is the use case you were thinking of?

adl1995:
@tboch I was actually asking for clarification on third point of the algorithm you mentioned. Where can I find the properties URL?

tboch:
Actually, the properties URL is the URL we pass as a parameter of class method `HipsSurveyProperties.fetch`
So, for example, if we start from the properties URL `'http://alasky.unistra.fr/DSS/DSS2Merged/properties'` and we can not find `hips_service_url` or `moc_access_url` , then `base_url` should return `'http://alasky.unistra.fr/DSS/DSS2Merged'`

[...]

PR #60, opened 2017-07-09
Update base_url property in HipsSurveyProperties class

adl1995:
This PR addresses issue #53.

### DIFF
--- a/hips/tiles/surveys.py
+++ b/hips/tiles/surveys.py
@@ -69,16 +69,18 @@ class HipsSurveyProperties:
 
         with urllib.request.urlopen(url) as response:
             text = response.read().decode('utf-8')
-        return cls.parse(text)
+        return cls.parse(text, url)
 
     @classmethod
-    def parse(cls, text: str) -> 'HipsSurveyProperties':
+    def parse(cls, text: str, url: str = None) -> 'HipsSurveyProperties':
         """Parse HiPS survey description text (`HipsSurveyProperties`).
 
         Parameters
         ----------
         text : str
             Text containing HiPS survey properties
+        url : str
+            Properties URL of HiPS
         """
         data = OrderedDict()
         for line in text.split('\n'):
@@ -91,6 +93,9 @@ class HipsSurveyProperties:
             except ValueError:
                 # Skip bad lines (silently, might not be a good idea to do this)
                 continue
+
+        if url is not None:
+            data['properties_url'] = url.rsplit('/', 1)[0]
 
         return cls(data)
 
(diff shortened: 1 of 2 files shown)


adl1995:
@tboch I just made some changes. Please review.

hips/tiles/surveys.py
```diff
@@ -127,7 +132,12 @@ def tile_format(self) -> str:
     @property
     def base_url(self) -> str:
         """HiPS access url"""
-        return self.data['moc_access_url'].rsplit('/', 1)[0]
+        try:
+            return self.data['hips_service_url']
+        except IndexError:
```
tboch: are you sure you ever enter in this except branch?
If the key `hips_service_url` is missing, isn't a `KeyError` raised rather than a `IndexError`?

hips/tiles/surveys.py
```diff
@@ -127,7 +132,12 @@ def tile_format(self) -> str:
     @property
     def base_url(self) -> str:
```
tboch: Do we have some test cases for the different cases? (base URL retrieved from hips_service_url, from moc_access_url or directly from the properties URL)

hips/tiles/surveys.py
```diff
@@ -127,7 +132,12 @@ def tile_format(self) -> str:
...
         """HiPS access url"""
-        return self.data['moc_access_url'].rsplit('/', 1)[0]
+        try:
+            return self.data['hips_service_url']
+        except IndexError:
+            return self.data['moc_access_url'].rsplit('/', 1)[0]
+        except:
```
tboch: 2 comments:

- I think this should be nested one level deeper
- what should we do in case we reach the last fallback and still don't get the base URL. Return None, raise an exception? @cdeil , what is the most Pythonic/Astropynic way of handling this?

cdeil: I would say: if in doubt, raise an exception, because it's better to fail quickly and loudly than later in weird ways if someone calls this and doesn't handle the error case.

adl1995: Is this the correct way to handle this?

```
    def base_url(self) -> str:
        """HiPS access url"""
        try:
            return self.data['hips_service_url']
        except KeyError:
            return self.data['moc_access_url'].rsplit('/', 1)[0]
        except KeyError:
            return self.data['properties_url']
        except:
            return ValueError('URL does not exist!')

```

I'm getting this warning in PyCharm:

> Exception class 'KeyError' has already been caught

hips/tiles/surveys.py
adl1995: @tboch Can you please list a survey which is missing both `hips_service_url` and `moc_access_url`.

tboch: Try to load properties from http://alasky.u-strasbg.fr/DSS/DSS2-NIR/properties for instance

hips/tiles/surveys.py
tboch: I would have done something like this:
```python
    def base_url(self) -> str:
        """HiPS access url"""
        try:
            return self.data['hips_service_url']
        except KeyError:
            try:
                return self.data['moc_access_url'].rsplit('/', 1)[0]
            except KeyError:
                try:
                    return self.data['properties_url']
                except:
                    return ValueError('URL does not exist!')
```

I agree it's a bit ugly

adl1995:
I made another commit updating the `try-except` block and adding a test case.

tboch:
Thanks!
PR merged.

Issue #62, opened 2017-07-09
Drawing JPG, PNG tiles


adl1995:
For drawing the higher dimensional tiles such as JPG and PNG, the empty image's shape in `draw_sky_image` would have to be altered. This `shape` currently comes from `WCSGeometry` class. However, this is always 2 dimensional. To handle multi-dimentional tiles, I'm thinking of adding them manually, i.e. 3 extra dimensions in case of JPG tiles and 4 extra dimensions in case of PNG tiles.

@cdeil @tboch If you have an alternative / better suggestion, please share.

cdeil:
> For drawing the higher dimensional tiles such as JPG and PNG, the empty image's shape in draw_sky_image would have to be altered. This shape currently comes from WCSGeometry class. However, this is always 2 dimensional.

For drawing RGB images, two different shapes are needed in different places of the code:
* The 3-dim shape of the ``data`` -- used e.g. when creating the empty sky image array at the start of the drawing.
* The 2-dim shape, with the third dimension dropped -- used e.g. to compute the transform (I assume scikit-image wants 2D here and can't handle 3D directly, but I'm not sure, you should try).

> To handle multi-dimentional tiles, I'm thinking of adding them manually, i.e. 3 extra dimensions in case of JPG tiles and 4 extra dimensions in case of PNG tiles.

I don't know how to handle this best yet, and I have no idea what you mean by "adding them manually".
My suggestion is that you start a PR, start by adding a (failing) test case  and then try to do minimal changes to the existing code to make it work for RGB images, asking for help on the PR as you go.

The core question is if the scikit-image warp function can process 3-dim or not.
* If yes, then maybe you just need to make the 2-dim shape when computing the transform (dropping the RGB dimension)?
* If no, then probably you need to do a Python loop over the RGB channels, and draw one after the other.

@adl1995 - Does this help you get started?

cdeil:
Implemented by @adl1995  in #63

PR #63, opened 2017-07-10
Add support for RGB tiles

adl1995:
As discussed in issue #62, I have added a test case of JPG tiles, which is passing. However, I haven't added a test case for PNG tiles, because the current survey only support FITS, JPG tiles.

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -18,8 +18,10 @@ __doctest_skip__ = [
     'compute_matching_hips_order',
 ]
 
+
 # TODO: Fix type annotation issue
-def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any], hips_survey: HipsSurveyProperties) -> np.ndarray:
+def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any],
+                   hips_survey: HipsSurveyProperties, tile_format: str) -> np.ndarray:
     """Draw sky image using the simple and quick method.
 
     Parameters
@@ -30,13 +32,21 @@ def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any], 
         A list of HipsTile
     hips_survey : `~hips.HipsSurveyProperties`
         HiPS survey properties
+    tile_format : `str`
+        Format of HiPS tile
 
     Returns
     -------
     np.ndarray
         Returns a numpy array containing all HiPS tiles projected onto it
     """
-    image = np.zeros(geometry.shape)
+    if tile_format == 'jpg':
+        shape = (geometry.shape.width, geometry.shape.height, 3)
+    elif tile_format == 'png':
+        shape = (geometry.shape.width, geometry.shape.height, 4)
+    else:
+        shape = (geometry.shape.width, geometry.shape.height)
+    image = np.zeros(shape)
     for tile in tiles:
         painter = SimpleTilePainter(geometry, hips_survey, tile)
         image += painter.warp_image()
@@ -45,7 +55,6 @@ def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any], 
 
 class SimpleTilePainter:
     """Paint a single tile using a simple projective transformation method.
-
     The algorithm implemented is described here: :ref:`drawing_algo`.
 
     Parameters
@@ -73,6 +82,7 @@ class SimpleTilePainter:
              [0, width - 1],
              [0, 0]],
         )
+
     @property
     def projection(self) -> ProjectiveTransform:
         """Estimate projective transformation on a HiPS tile"""
@@ -93,7 +103,8 @@ class SimpleTilePainter:
         )
 
 
-def fetch_tiles(healpix_pixel_indices: np.ndarray, order: int, hips_survey: HipsSurveyProperties) -> 'HipsTile':
+def fetch_tiles(healpix_pixel_indices: np.ndarray, order: int,
+                hips_survey: HipsSurveyProperties, tile_format: str) -> 'HipsTile':
     """Fetch HiPS tiles from a remote URL.
 
     Parameters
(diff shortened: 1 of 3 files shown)


tboch:
Could you use the 2MASS6XH sample in hips-extra, which has PNG tiles?

adl1995:
@tboch The [PNG tiles](https://github.com/hipspy/hips-extra/tree/master/datasets/samples/2MASS6XH/Norder6/Dir0) are already available in hips-extra. I was reluctant to add the test case as it would require a separate survey, the current one does not support PNG tiles.

I think a separate function `get_test_tiles_png` would server the purpose, or would it be better to take the tile path as a parameter?

tboch:
I have no strong feeling about how to test this. But if the library supports PNG, we should add a test using PNG tiles.

adl1995:
I just added a test case for PNG tiles, however, it is currently failing. The pixel values sum to `0.0`. I suspect this is because it doesn't fall in the sky image.

cdeil:
You have committed changes to `astropy_helpers`:
https://github.com/hipspy/hips/pull/63/files
Please get rid of them (and never commit changes to `astropy_helpers` in the future).

hips/draw/simple.py
```diff
@@ -30,13 +32,21 @@ def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any],
         A list of HipsTile
     hips_survey : `~hips.HipsSurveyProperties`
         HiPS survey properties
+    tile_format : `str`
```
cdeil: Why pass this around separately?
It's a property of https://hips.readthedocs.io/en/latest/api/hips.HipsSurveyProperties.html#hips.HipsSurveyProperties.tile_format , no?

hips/draw/simple.py
```diff
@@ -30,13 +32,21 @@ def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any],
...
+        Format of HiPS tile
 
     Returns
     -------
     np.ndarray
         Returns a numpy array containing all HiPS tiles projected onto it
     """
-    image = np.zeros(geometry.shape)
+    if tile_format == 'jpg':
+        image = np.zeros((geometry.shape.width, geometry.shape.height, 3))
```
cdeil: Nitpick comment: I'd suggest to just compute a local `shape` variable here in the if / else block, and then to put `image = np.zeros(shape)` below, i.e. not duplicate the `np.zeros` code multiple times.

hips/draw/tests/test_simple.py
```diff
@@ -4,45 +4,53 @@
...
 
-    tile1 = HipsTile.read(
-        meta=HipsTileMeta(order=3, ipix=450, file_format='fits', frame=hips_survey.astropy_frame),
-        full_path=get_hips_extra_file('datasets/samples/DSS2Red/Norder3/Dir0/Npix450.fits'),
-    )
+    tiles = []
+    for ipix in ipix_list:
+        tiles.append(HipsTile.read(
+            meta=HipsTileMeta(order=order, ipix=ipix, file_format=file_format, frame=hips_survey.astropy_frame),
+            full_path=get_hips_extra_file('datasets/samples/' + survey + '/Norder' + str(order) + '/Dir0/Npix' + str(ipix) + '.' + file_format),
```
cdeil: Shouldn't building this string be a property or method somewhere, e.g. on `HipsTileMeta`?
```
survey + '/Norder' + str(order) + '/Dir0/Npix' + str(ipix) + '.' + file_format
```
Otherwise we duplicate that code 100 times in our repo.

hips/draw/tests/test_simple.py
```diff
@@ -51,9 +59,10 @@ def test_make_sky_image():
     url = 'http://alasky.unistra.fr/DSS/DSS2Merged/properties'
     hips_survey = HipsSurveyProperties.fetch(url)
     geometry = make_test_wcs_geometry(case=2)
-    data = make_sky_image(geometry, hips_survey)
+    data = make_sky_image(geometry=geometry, hips_survey=hips_survey, tile_format='fits')
     assert data.shape == geometry.shape
     assert data.dtype == np.float64
+    print(np.max(data))
```
cdeil: Remove print, and maybe instead add assert?

cdeil:
> I just added a test case for PNG tiles, however, it is currently failing. The pixel values sum to 0.0. I suspect this is because it doesn't fall in the sky image.

Can you pick another survey that has PNG and data at least for part of the sky image?

tboch:
>      Can you pick another survey that has PNG and data at least for part of the sky image?

I'd suggest using HiPS [P/AKARI/FIS/Color](http://alasky.u-strasbg.fr/AKARI-FIS/ColorLSN60/) that has PNG and data in the galactic plane.

hips/draw/simple.py
adl1995: @cdeil I don't think I can use that. The properties file could contain multiple formats such as 'jpg fits'.

cdeil: Ah, in that case: OK to add a separate parameter, at least for now.

cdeil:
@adl1995 - I merged https://github.com/hipspy/hips-extra/pull/3 just now.
Does it work if you use tiles from that survey?

cdeil:
@adl1995 - Maybe add a section on making an RGB color image in the high-level docs with a nice example? https://hips.readthedocs.io/en/latest/getting_started.html#more-advanced-examples

The HiPS paper uses "HiPS of the Spitzer GLIMPSE 360 survey" in Figure 4, which looks nice.

adl1995:
@cdeil I just made some updates. For the example on getting started page, should I duplicate all the steps from above or only the ones that need changing?

hips/tiles/surveys.py
```diff
@@ -150,6 +150,20 @@ def tile_access_url(self, order: int, ipix: int) -> str:
...
+        """Tile access URL
+
+        Parameters
+        ----------
+        order : int
+            HiPS order
+        ipix : int
+            Index of the HiPS tile
+        tile_format : str
+            HiPS tile URL
```
cdeil: Parameter description isn't correct for `tile_format`.

Also: maybe add an `Examples` section with one example?

cdeil:
> For the example on getting started page, should I duplicate all the steps from above or only the ones that need changing?

As you like. I think I would re-use the `geometry`, but otherwise it's new.

Things to explain there:
* That HiPS also has JPEG and PNG, pointing out one example for each
* What the return data array is (it's shape, and that it's float even though int pixel data wend in from the tiles).
* End the example with how RGB images can be plotted with matplotlib / wcsaxes and saved to FITS or PNG / JPEG? (i.e. tell users a bit how they can work with the image they generate)

adl1995:
@cdeil Please merge this PR. I will make a separate PR for documentation changes.

cdeil:
OK, merging now.

Issue #57, opened 2017-07-06
Reformating SimpleTilePainter class


adl1995:
As discussed during our telecon on July 5th, the `SimpleTilePainter` class needs to be reformated to exactly outline the steps mentioned in the [drawing algorithm](http://hips.readthedocs.io/en/latest/drawing_algo.html). The current suggestions are outlined below:

- Move tile list computation part within `SimpleTilePainter`
- `SimpleTilePainter` should be passed in a list of tiles instead of a single tile
- Methods should be converted into properties and there should be a method named `run` which executes all steps in order
- Add unit tests for all the intermediate values computed in `SimpleTilePainter` class

@tboch @cdeil - If you have any other suggestions, please share.

cdeil:
> SimpleTilePainter should be passed in a list of tiles instead of a single tile

I would suggest that the painter is responsible for all steps, even tile list computation and fetching.
Internally it would eventually create a `HipsTileCache` and call it to fetch and store the tiles, for now it calls out to the compute tiles needed and fetch tiles needed functions.

I.e. it gets initialised just with the user spec of geometry and which HiPS to use (at the moment represented by a `HipsSurveyProperties` object, and then runs all the steps from the drawing algorithm description.

cdeil:
Done by @adl1995 in #66 .

PR #66, opened 2017-07-11
Reformat SimpleTilePainter class

adl1995:
As discussed in issue #57, I have reformatted the `SimpleTilePainter` class. I'm aware this still needs some changes, but this will provide us with a working ground. For now, I have commented out the `test_draw_sky_image` function as it is not currently usable.

@cdeil @tboch Please provide feedback on what needs to be changed.

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -1,61 +1,30 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """HiPS tile drawing -- simple method."""
-from typing import Generator, Any
 import numpy as np
+from typing import Tuple
 from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
 from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta
-from ..utils import WCSGeometry, compute_healpix_pixel_indices
+from ..utils import WCSGeometry, compute_healpix_pixel_indices, get_hips_order_for_resolution
+
 
 __all__ = [
-    'draw_sky_image',
     'make_sky_image',
     'SimpleTilePainter',
-    'compute_matching_hips_order'
+
 ]
 
 __doctest_skip__ = [
-    'compute_matching_hips_order',
+    'SimpleTilePainter'
 ]
 
 
-# TODO: Fix type annotation issue
-def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any],
-                   hips_survey: HipsSurveyProperties, tile_format: str) -> np.ndarray:
+class SimpleTilePainter:
     """Draw sky image using the simple and quick method.
 
-    Parameters
-    ----------
-    geometry : `~hips.utils.WCSGeometry`
-        An object of WCSGeometry
-    tiles : List[HipsTile]
-        A list of HipsTile
-    hips_survey : `~hips.HipsSurveyProperties`
-        HiPS survey properties
-    tile_format : `str`
-        Format of HiPS tile
-
-    Returns
-    -------
-    np.ndarray
-        Returns a numpy array containing all HiPS tiles projected onto it
-    """
-    if tile_format == 'jpg':
-        shape = (geometry.shape.height, geometry.shape.width, 3)
-    elif tile_format == 'png':
-        shape = (geometry.shape.height, geometry.shape.width, 4)
-    else:
-        shape = (geometry.shape.height, geometry.shape.width)
-    image = np.zeros(shape)
-    for tile in tiles:
-        painter = SimpleTilePainter(geometry, hips_survey, tile)
-        image += painter.warp_image()
-    return image
-
-
-class SimpleTilePainter:
-    """Paint a single tile using a simple projective transformation method.
-    The algorithm implemented is described here: :ref:`drawing_algo`.
+    Paint HiPS tiles onto an all-sky image using a simple projective
+    transformation method. The algorithm implemented is described
+    here: :ref:`drawing_algo`.
 
     Parameters
     ----------
@@ -63,95 +32,14 @@ class SimpleTilePainter:
         An object of WCSGeometry
     hips_survey : `~hips.HipsSurveyProperties`
         HiPS survey properties
-    tile : `HipsTile`
-       An object of HipsTile
-    """
-
-    def __init__(self, geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile: HipsTile) -> None:
-        self.geometry = geometry
-        self.hips_survey = hips_survey
-        self.tile = tile
-
-    @property
-    def dst(self) -> np.ndarray:
-        """Destination array for projective transform"""
-        width = self.hips_survey.tile_width
-        return np.array(
-            [[width - 1, 0],
-             [width - 1, width - 1],
-             [0, width - 1],
-             [0, 0]],
-        )
-
-    @property
-    def projection(self) -> ProjectiveTransform:
-        """Estimate projective transformation on a HiPS tile"""
-        corners = self.tile.meta.skycoord_corners.to_pixel(self.geometry.wcs)
-        src = np.array(corners).T.reshape((4, 2))
-        dst = self.dst
-        pt = ProjectiveTransform()
-        pt.estimate(src, dst)
-        return pt
-
-    def warp_image(self) -> np.ndarray:
-        """Warp a HiPS tile and a sky image"""
-        return warp(
-            self.tile.data,
-            self.projection,
-            output_shape=self.geometry.shape,
-            preserve_range=True,
-        )
-
-
-def fetch_tiles(healpix_pixel_indices: np.ndarray, order: int,
-                hips_survey: HipsSurveyProperties, tile_format: str) -> 'HipsTile':
-    """Fetch HiPS tiles from a remote URL.
-
-    Parameters
-    ----------
-    healpix_pixel_indices : np.ndarray
-        A list of HEALPix pixel indices
-    order : int
-        Order of the HEALPix map
-    hips_survey : HipsSurveyProperties
-        An object of HipsSurveyProperties
     tile_format : `str`
         Format of HiPS tile
-
-    Returns
-    -------
-    'HipsTile'
-        Returns an object of HipsTile
-    """
-    for healpix_pixel_index in healpix_pixel_indices:
-        tile_meta = HipsTileMeta(
-            order=order,
-            ipix=healpix_pixel_index,
-            frame=hips_survey.astropy_frame,
-            file_format=tile_format,
-        )
-        tile = HipsTile.fetch(tile_meta, hips_survey.tile_access_url(order=order, ipix=healpix_pixel_index) + tile_meta.filename)
-        yield tile
-
-
-def compute_matching_hips_order(geometry: WCSGeometry, hips_survey: HipsSurveyProperties) -> int:
-    """Compute HiPS tile order matching a given image pixel size.
-
-    Parameters
-    ----------
-    geometry : WCSGeometry
-        Geometry of the output image
-    hips_survey : HipsSurveyProperties
-        An object of HipsSurveyProperties
-
-    Returns
-    -------
-    'int'
-        Returns HiPS order
 
     Examples
     --------
-    >>> from hips.draw import compute_matching_hips_order
+    >>> from hips.utils import WCSGeometry
+    >>> from hips.draw import SimpleTilePainter
+    >>> from hips.tiles import HipsSurveyProperties
     >>> from astropy.coordinates import SkyCoord
     >>> url = 'http://alasky.unistra.fr/DSS/DSS2Merged/properties'
     >>> hips_survey = HipsSurveyProperties.fetch(url)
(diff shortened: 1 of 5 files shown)


hips/draw/simple.py
```diff
@@ -65,12 +30,64 @@ class SimpleTilePainter:
...
+        """
+        # Sky image angular resolution (pixel size in degree)
+        resolution = np.min(proj_plane_pixel_scales(self.geometry.wcs))
+        desired_order = self._get_hips_order_for_resolution(self.hips_survey.tile_width, resolution)
+        # Return the desired order, or the highest resolution available.
+        # Note that HiPS never has resolution less than 3,
+        # and that limit is handled in _get_hips_order_for_resolution
+        return np.min([desired_order, self.hips_survey.hips_order])
+
+    def _get_hips_order_for_resolution(self, tile_width, resolution):
```
cdeil: This function is not accessing `self`. Please either make it a `@staticmethod` or move it to a standalone function above this class, or to `hips/utils/healpix.py`. All of these options seem roughly as well as the other to me, your choice.

hips/draw/simple.py
```diff
@@ -65,12 +30,64 @@ class SimpleTilePainter:
...
+        for candidate_tile_order in range(3, 29 + 1):
+            tile_resolution = np.sqrt(full_sphere_area / 12 / 4 ** (candidate_tile_order + tile_order))
+            # Finding the smaller tile order with a resolution equal or better than geometric resolution
+            if tile_resolution <= resolution:
+                break
+
+        return candidate_tile_order
+
+    @property
+    def tile_indices(self):
```
cdeil: Please always add at least one-line docstrings to every property, function, class, method, ... everything that appears in the public API docs. The most important thing to mention is what it is and what type it is.

hips/draw/simple.py
```diff
@@ -83,108 +100,53 @@ def dst(self) -> np.ndarray:
...
+                ipix=healpix_pixel_index,
+                frame=self.hips_survey.astropy_frame,
+                file_format=self.tile_format,
+            )
+            tile = HipsTile.fetch(tile_meta, self.hips_survey.tile_access_url(order=self.hips_order, ipix=healpix_pixel_index) + tile_meta.filename)
+            yield tile
+
+    @property
+    def shape(self) -> tuple:
+        """Shape of the output image"""
```
cdeil: Would this be a good place to document what the shape will be for the three format cases?
Or can you add this info what shape computed images have somewhere else in the docs?

hips/draw/simple.py
```diff
@@ -83,108 +100,53 @@ def dst(self) -> np.ndarray:
...
-    for candidate_tile_order in range(3, 29 + 1):
-        tile_resolution = np.sqrt(full_sphere_area / 12 / 4 ** (candidate_tile_order + tile_order))
-        # Finding the smaller tile order with a resolution equal or better than geometric resolution
-        if tile_resolution <= resolution:
-            break
-
-    return candidate_tile_order
+    def run(self) -> np.ndarray:
+        """Run all steps of naive algorithm"""
+        tiles = self.fetch_tiles()
```
cdeil: Maybe store the `tiles` on `self` here?
Could be useful for tests, or power-users, to have access to the tiles after running the whole procedure, no?
(it's just a question, I'm not sure yet how to best structure the code here)

hips/draw/simple.py
```diff
@@ -83,108 +100,53 @@ def dst(self) -> np.ndarray:
...
+        image = np.zeros(self.shape)
+        for tile in tiles:
+            image += self.warp_image(tile)
+        return image
```
cdeil: Same for `image`, I would suggest to store it on `self`.
And I would suggest that the `run` method doesn't return it, but that users do:
```
painter = SimpleTilePainter(...)
painter.run()
# and then can access results you're interested in, for now mostly
painter.image
# but later also other things
painter.tiles
painter.info -> could be a dict where we store run-time, memory used, basic infos like number of tiles fetched, ... things that are useful to print in a report of what the algorithm did after running it.
```

hips/draw/tests/test_simple.py
```diff
@@ -25,32 +25,32 @@ def get_test_tiles(file_format, survey, order, ipix_list):
...
+#     url = 'https://raw.githubusercontent.com/hipspy/hips-extra/master/datasets/samples/' + pars['survey'] + '/properties'
+#     hips_survey = HipsSurveyProperties.fetch(url)
+#
+#     data = draw_sky_image(geometry=geometry, tiles=tiles, hips_survey=hips_survey, tile_format=pars['file_format'])
+#
+#     assert data.shape == pars['shape']
+#     assert data.dtype == np.float64
+#     assert_allclose(np.sum(data), pars['data_sum'])
+#     assert_allclose(data[400, 500], pars['data_1'])
+#     assert_allclose(data[400, 501], pars['data_2'])
```
cdeil: What's the plan with all this commented-out code?
We should make a decision and either simply remove it or refactor it into equivalent new code.
I don't want to merge as long as there's still this old commented-cod around.
@adl1995 - What's your suggestion? Is this still useful or is superfluous or already refactored elsewhere now?

adl1995: @cdeil I think this can be updated according to the current code. However, in the current commented-out code the `tiles` are passed in manually, but the updated code computes them after calling `healpix_pixel_indices`.

So, to use this as-it-is I would need to add the missing tiles to `hips-extra` so the test doesn't fail. One other option could be to pass in an optional parameter for `tiles`, so only the provided tiles will be draw. Do you have any opinion on this?

cdeil: I think `run` should execute all steps, i.e. basically we end up with this:
```python
def make_sky_image(geometry, survey, maybe_other_config):
    painter = SimpleTilePainter(geometry, survey, maybe_other_config)
    painter.run() # execute all steps from the high-level docs page describing simple tile painting, adjust algorithm description if needed
    return painter.image
```
i.e. `make_sky_image` is a simple and small wrapper of `SimpleTilePainter`.
the `SimpleTilePainter` takes all or most infos in `__init__` and then stores intermediate things it computes on `self` while running, and at the end has all infos available as data members after `run` was executed.

I'm not sure if this answers your question here? Probably we need to discuss a bit in the telcon?

adl1995:
@cdeil I made some changes. Although, I'm not sure if the `make_sky_image` function is required now, as it basically just calls `SimpleTilePainter.run()`. Also, I couldn't add tests for png tiles because they are not available in `hips-extra` (the ones that fall on the galactic center).

cdeil:
I'll have a look now.
My suggestion would be to keep the "make_sky_image" function, as the very high level, very simple end user interface. Once we have different tile fetching / caching and drawing methods, the code in `make_sky_image` would again grow to 10 or 20 lines of configuring / calling the lower-level routines via a uniform and simple to configure interface (e.g. a `draw_method="simple"` or `draw_method="fancy"` method.

`make_sky_image` should also return a result object that offers more info / data members than just the numpy array of pixels we currently return. 

Makes sense?

cdeil:
> Also, I couldn't add tests for png tiles because they are not available in hips-extra (the ones that fall on the galactic center).

Can you make a PR adding the tiles you need yourself? If yes, please do. If no, please file an issue and I or Thomas will do it.

hips/draw/simple.py
```diff
@@ -83,111 +87,60 @@ def dst(self) -> np.ndarray:
...
+        self.tiles = self.fetch_tiles()
+        self.image = np.zeros(self.shape)
+        for tile in self.tiles:
+            self.image += self.warp_image(tile)
 
 
-def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile_format: str) -> np.ndarray:
+def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile_format: str) -> 'SimpleTilePainter':
     """Make sky image: fetch tiles and draw.
     The example for this can be found on the :ref:`gs` page.
```
cdeil: Add an empty line after the summary line of the `make_sky_image` docstring (and always, for any docstring).

hips/draw/simple.py
```diff
@@ -65,12 +29,52 @@ class SimpleTilePainter:
         HiPS survey properties
     tile : `HipsTile`
```
cdeil: The `SimpleTilePainter` no longer takes a `HipsTile` in `__init__`.

Also the description above is no longer correct, it still says "paint a single tile"

Update it's docstring!

hips/draw/simple.py
```diff
@@ -83,111 +87,60 @@ def dst(self) -> np.ndarray:
...
+    def fetch_tiles(self) -> 'HipsTile':
+        """Fetch HiPS tiles from a remote URL"""
+        for healpix_pixel_index in self.tile_indices:
+            tile_meta = HipsTileMeta(
+                order=self.hips_order,
+                ipix=healpix_pixel_index,
+                frame=self.hips_survey.astropy_frame,
+                file_format=self.tile_format,
+            )
+            tile = HipsTile.fetch(tile_meta, self.hips_survey.tile_access_url(order=self.hips_order, ipix=healpix_pixel_index) + tile_meta.filename)
```
cdeil: This line is way too long. Move the computation of the URL on the line before and then pass it here.

hips/draw/simple.py
```diff
@@ -65,12 +29,52 @@ class SimpleTilePainter:
...
         self.geometry = geometry
         self.hips_survey = hips_survey
-        self.tile = tile
+        self.tile_format = tile_format
+
+    @property
+    def hips_order(self) -> int:
+        """Compute HiPS tile order matching a given image pixel size.
+
+        Examples
```
cdeil: This is a weird place to have the only example for this class. Move it to the class docstring (and maybe change it a bit, to show what properties are available after calling `run`.

hips/draw/simple.py
```diff
@@ -83,111 +87,60 @@ def dst(self) -> np.ndarray:
...
+                order=self.hips_order,
+                ipix=healpix_pixel_index,
+                frame=self.hips_survey.astropy_frame,
+                file_format=self.tile_format,
+            )
+            tile = HipsTile.fetch(tile_meta, self.hips_survey.tile_access_url(order=self.hips_order, ipix=healpix_pixel_index) + tile_meta.filename)
+            yield tile
+
+    @property
+    def shape(self) -> tuple:
```
cdeil: Suggest to rename to more explicit `output_image_shape`, and to make a more precise type annotation (I think it's `Tuple[int]`

The description below is not quite correct. You say "four extra dimensions".

What we really have is three dimensions for RGB images, and there are four "channels" for the case of PNG:
```
In [1]: import numpy as np

In [2]: data = np.zeros((100, 200, 4))

In [3]: data.shape
Out[3]: (100, 200, 4)

In [4]: data.ndim
Out[4]: 3
```

My suggestion here would be that you state that we follow the same axis order and coordinate conventions that are used by others for grayscale and RGB images:
http://scikit-image.org/docs/stable/user_guide/numpy_images.html#coordinate-conventions
and then give examples (like they do on that page, repeating the info relevant for uw) showing the shape.

hips/draw/simple.py
```diff
@@ -83,111 +87,60 @@ def dst(self) -> np.ndarray:
...
+        """Shape of the output image
+
+        The shape will be two dimensional in case of FITS file format,
+        extra three dimensions (RGB) will be added in case of JPG,
+        and four extra dimensions (RGBA) in case of PNG tile.
+        """
+        if self.tile_format == 'jpg':
+            return (self.geometry.shape.height, self.geometry.shape.width, 3)
```
cdeil: remove redundant parentheses.

hips/draw/simple.py
```diff
@@ -202,18 +155,12 @@ def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties, til
...
+    painter : `SimpleTilePainter`
+        An object of SimpleTilePainter class
     """
-    order = compute_matching_hips_order(geometry, hips_survey)
-    healpix_pixel_indices = compute_healpix_pixel_indices(
-        wcs_geometry=geometry,
-        order=order,
-        healpix_frame=hips_survey.astropy_frame,
-    )
     # TODO: this isn't a good API. Will become better when we have a cache.
```
cdeil: Remove this old TODO comment.

hips/draw/simple.py
```diff
@@ -202,18 +155,12 @@ def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties, til
 
     Returns
     -------
-    data : `~numpy.ndarray`
-        Output image pixels
+    painter : `SimpleTilePainter`
```
cdeil: For now, I would suggest to only return `data` like we did before.

I think the Painter should be more or less an implementation detail, and we should introduce a results class in the future that is then uniform for different painters and presents the result info in a better way than a painter class with all the methods mixed in.

hips/draw/simple.py
```diff
@@ -83,111 +87,60 @@ def dst(self) -> np.ndarray:
...
-        tile_resolution = np.sqrt(full_sphere_area / 12 / 4 ** (candidate_tile_order + tile_order))
-        # Finding the smaller tile order with a resolution equal or better than geometric resolution
-        if tile_resolution <= resolution:
-            break
-
-    return candidate_tile_order
+    def run(self) -> None:
+        """Run all steps of naive algorithm"""
+        self.tiles = self.fetch_tiles()
+        self.image = np.zeros(self.shape)
```
cdeil: This should be changed to `self.image = self.draw_tiles()`, i.e. move those lines doing the drawing and stitching of the tiles to a method that can be executed and tested separately.

hips/draw/tests/test_simple.py
```diff
@@ -1,102 +1,79 @@
...
+        url = 'http://alasky.unistra.fr/DSS/DSS2Merged/properties'
+        cls.hips_survey = HipsSurveyProperties.fetch(url)
+        cls.geometry = WCSGeometry.create_simple(
+            skydir=SkyCoord(0, 0, unit='deg', frame='icrs'),
+            width=2000, height=1000, fov="3 deg",
+            coordsys='icrs', projection='AIT'
+        )
+        cls.simple_tile_painter = SimpleTilePainter(cls.geometry, cls.hips_survey, 'fits')
+
+    @remote_data
```
cdeil: At the moment you added `@remote_data` on every single method of `TestSimpleTilePainter`.
Please remove it from all the methods, and only put it on `TestSimpleTilePainter` instead.
(and then later we'll try to change the code / tests so that most tests don't require remote data, but not now.)

adl1995:
@cdeil I just made some changes.

cdeil:
Please always run `python setup.py build_docs` locally and resolve all warnings / errors:
https://travis-ci.org/hipspy/hips/jobs/253249006#L1158

hips/draw/tests/test_simple.py
```diff
@@ -1,102 +1,74 @@
...
-
-    return tiles
-
-
-draw_sky_image_pars = [
-    dict(file_format='fits', shape=(1000, 2000), survey='DSS2Red', data_1=2866.0101409848185,
-         data_2=2563.6916727348043, data_sum=4575235421.512643, order=3, ipix_list=[450, 451]),
-    dict(file_format='jpg', shape=(1000, 2000, 3), survey='DSS2Red', data_1=[13.040878, 13.040878, 13.040878],
-         data_2=[17.235874, 17.235874, 17.235874], data_sum=243177268.56158745, order=3, ipix_list=[450, 451]),
-    dict(file_format='png', shape=(1000, 2000, 4), survey='AKARI-FIS', data_1=[254., 254., 254., 255.],
```
cdeil: Why are you removing the PNG test here?

I looked at this earlier today, and I don't see why you couldn't just keep it!?
https://github.com/hipspy/hips-extra/issues/5#issuecomment-315030888

hips/draw/simple.py
```diff
@@ -1,76 +1,89 @@
...
+    >>> geometry = WCSGeometry.create_simple(
+    ...     skydir=SkyCoord(0, 0, unit='deg', frame='icrs'),
+    ...     width=2000, height=1000, fov="3 deg",
+    ...     coordsys='icrs', projection='AIT'
+    ... )
+    >>> simple_tile_painter = SimpleTilePainter(geometry, hips_survey, 'fits')
+    >>> simple_tile_painter.hips_order
+    7
+    >>> simple_tile_painter.run()
+    >>> simple_tile_painter.image[0]
```
cdeil: This is not the current output. It would be:
```
array([ 1866.7819988 ,  1865.42884071,  2494.00873715, ...,  1970.00283741,
        1998.08389507,  2045.58599973])
```
But I suggest you just do `painter.shape` instead of `painter.image[0]` here.

hips/draw/simple.py
```diff
@@ -1,76 +1,89 @@
...
+    >>> from hips.tiles import HipsSurveyProperties
+    >>> from astropy.coordinates import SkyCoord
+    >>> url = 'http://alasky.unistra.fr/DSS/DSS2Merged/properties'
+    >>> hips_survey = HipsSurveyProperties.fetch(url)
+    >>> geometry = WCSGeometry.create_simple(
+    ...     skydir=SkyCoord(0, 0, unit='deg', frame='icrs'),
+    ...     width=2000, height=1000, fov="3 deg",
+    ...     coordsys='icrs', projection='AIT'
+    ... )
+    >>> simple_tile_painter = SimpleTilePainter(geometry, hips_survey, 'fits')
```
cdeil: Rename `simple_tile_painter` -> `painter` to be shorter.

hips/draw/simple.py
```diff
@@ -1,76 +1,89 @@
...
 
-    def __init__(self, geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile: HipsTile) -> None:
+    def __init__(self, geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile_format: str) -> None:
         self.geometry = geometry
         self.hips_survey = hips_survey
-        self.tile = tile
+        self.tile_format = tile_format
+
+    @property
+    def hips_order(self) -> int:
```
cdeil: Suggest to rename this to the more specific term `draw_hips_order` to avoid confusion with e.g. the `hips_survey.hips_order`.

hips/draw/simple.py
```diff
@@ -83,112 +96,67 @@ def dst(self) -> np.ndarray:
...
         return warp(
-            self.tile.data,
-            self.projection,
+            tile.data,
+            self.projection(tile),
             output_shape=self.geometry.shape,
             preserve_range=True,
         )
 
+    def draw_tiles(self):
```
cdeil: Missing docstring on public method.

hips/draw/simple.py
```diff
@@ -83,112 +96,67 @@ def dst(self) -> np.ndarray:
...
+                frame=self.hips_survey.astropy_frame,
+                file_format=self.tile_format,
+            )
+            url = self.hips_survey.tile_access_url(order=self.hips_order, ipix=healpix_pixel_index) + tile_meta.filename
+            tile = HipsTile.fetch(tile_meta, url)
+            yield tile
+
+    @property
+    def shape(self) -> Tuple[int]:
+        """Shape of the output image
```
cdeil: Docstring rule: Always end summary docstring line with a `.`

hips/draw/simple.py
```diff
@@ -83,112 +96,67 @@ def dst(self) -> np.ndarray:
...
+        corners = tile.meta.skycoord_corners.to_pixel(self.geometry.wcs)
         src = np.array(corners).T.reshape((4, 2))
         dst = self.dst
         pt = ProjectiveTransform()
         pt.estimate(src, dst)
         return pt
 
-    def warp_image(self) -> np.ndarray:
+    def fetch_tiles(self) -> 'HipsTile':
+        """Fetch HiPS tiles from a remote URL"""
```
cdeil: From the docs it's not clear that this is a generator.
If the Type annotation doesn't work properly, then at least mention that this is a generator that returns `HipsTile` objects in the docstring.

hips/draw/simple.py
```diff
@@ -1,76 +1,89 @@
...
+    'SimpleTilePainter'
 ]
 
 
-# TODO: Fix type annotation issue
-def draw_sky_image(geometry: WCSGeometry, tiles: Generator[HipsTile, Any, Any],
-                   hips_survey: HipsSurveyProperties, tile_format: str) -> np.ndarray:
-    """Draw sky image using the simple and quick method.
+class SimpleTilePainter:
+    """Paint HiPS tiles onto an all-sky image using a simple projective
```
cdeil: Docstring rule: Always have a one-line summary docstring, optionally followed by an empty line and more text.

In this case, the docstring we had on `draw_sky_image` would be good for the `SimpleTilePainter` class IMO:
```
Draw sky image using the simple and quick method.
```

cdeil:
@adl1995 - I left some more small comments inline. Once those are adressed, and there's green light on travis ci (still the docs build: https://travis-ci.org/hipspy/hips/jobs/253258805#L1158), I'll merge.

cdeil:
@adl1995 - I wanted to start playing around with HiPS tile drawing a bit myself:
https://github.com/cdeil/hipsperiments/blob/master/rgb_tile_draw_issue.ipynb

The problem with the current class is that one cannot access the tiles any more after running the algorithm.

Can you please change the implementation for `painter.tiles` to return a list of tiles, instead of a being a generator.

Basically I would propose this coding pattern:
```
class Painter:
    def __init__(self):
        self._tiles = None

    @property
    def tiles(self):
        if self._tiles is None:
            self._tiles = list(self._fetch_tiles())  # the method you have now, just with a leading underscore
        return self._tiles
```

This way, the painter fetches the tiles only once, and puts them in his pocket in case someone (like me) wants to inspect them later.

@adl1995 - is it clear? do you want to implement it or should I just take over here?

adl1995:
@cdeil I made some updates to the code and also added a test case for PNG tiles. The Travis CI is passing now. Please check.

cdeil:
@adl1995 - Thanks!

I'm merging this now.

PS: I'll do some cleanup and coding on `hips` in the next ~ 2 hours, to polish a bit for the upcoming v0.1 release.

Issue #65, opened 2017-07-11
width and height inverted in WCSGeometry.create_simple ?


tboch:
Following the instructions of the _Getting started_, the dimension of the result image is not as expected: it is **1000 pixels wide and 2000 pixels tall** whereas passed parameters are `width=2000, height=1000`

cdeil:
@tboch - Good catch!

Should be fixed now via #67 .

PR #67, opened 2017-07-12
Fix tile width height issue

adl1995:
As pointed out by @tboch in issue #65, the `width` and `height` of the HiPS tiles was inverted. This PR will fix the issue.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -41,7 +41,7 @@ To make a sky image with the `hips` package, follow the following three steps:
 
     from hips.draw import make_sky_image
 
-    data = make_sky_image(geometry, hips_survey)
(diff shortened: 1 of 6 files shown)


cdeil:
I see at http://scikit-image.org/docs/stable/user_guide/numpy_images.html that indeed they have the `(row, col)`, i.e. `(height, width)` axis order:

screen shot 2017-07-12 at 08 59 47

So yes, this PR looks good to me.

Please fix the high-level docs example here (add `tile_format`) and show that it works by pasting the image you get here from that example (you can drag and drop PNGs into Github comment fields).

adl1995:
@cdeil The high level docs code is now working. Here is the output:

sky

cdeil:
Looks good to me. Thanks!

PR #68, opened 2017-07-12
Add plot for image in high level docs

adl1995:
This PR is not ready to be merged. It just shows that the plot for jpg is not in RGB. So, there might be an issue in drawing.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -54,7 +54,9 @@ Now you can then save the sky image to local disk e.g. FITS file format::
 
 or plot and analyse the sky image however you like.
 
-TODO: show how to plot the image with `astropy.visualization.wcsaxes`.
+If you execute the example above, you will get this sky image which was plotted using `astropy.visualization.wcsaxes`
+
+.. plot:: plot_fits.py
 
 HiPS data
 ---------
--- a/docs/plot_fits.py
+++ b/docs/plot_fits.py
@@ -1,0 +1,20 @@
+"""This script plots the all-sky image following the example on getting started page"""
+import matplotlib.pyplot as plt
+from astropy.coordinates import SkyCoord
+from astropy.visualization.mpl_normalize import simple_norm
+
+from hips import HipsSurveyProperties
+from hips import make_sky_image
+from hips.utils import WCSGeometry
+import numpy as np
+url = 'http://alasky.unistra.fr/DSS/DSS2Merged/properties'
+hips_survey = HipsSurveyProperties.fetch(url)
+geometry = WCSGeometry.create_simple(
+    skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
+    width=2000, height=1000, fov="3 deg",
+    coordsys='galactic', projection='AIT'
+)
+data = make_sky_image(geometry=geometry, hips_survey=hips_survey, tile_format='fits')
+ax = plt.subplot(projection=geometry.wcs)
+norm = simple_norm(data, 'sqrt')
+ax.imshow(data, origin='lower', norm=norm, cmap='gray')
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,6 +11,7 @@ show-response = 1
 minversion = 3.0
 norecursedirs = build docs/_build
 doctest_plus = enabled
+doctest_norecursedirs = docs/plot_fits.py
 
 [ah_bootstrap]
 auto_use = True


docs/plot.py
```diff
@@ -0,0 +1,17 @@
+import matplotlib.pyplot as plt
```
cdeil: Always add a one-line docstring at the top of scripts saying what it is.

docs/plot.py
```diff
@@ -0,0 +1,17 @@
...
+from astropy.coordinates import SkyCoord
+
+from hips.draw.simple import make_sky_image
```
cdeil: Import from the top-level `hips` namespace here and on the line below.
The basic idea is to always import from the location where something appears in the API docs, i.e. where users might find it if they look at the docs.

docs/getting_started.rst
```diff
@@ -56,6 +56,8 @@ or plot and analyse the sky image however you like.
 
 TODO: show how to plot the image with `astropy.visualization.wcsaxes`.
```
cdeil: Edit this line and put some useful text.

docs/plot.py
```diff
@@ -0,0 +1,17 @@
...
+from hips.utils import WCSGeometry
+
+url = 'http://alasky.unistra.fr/DSS/DSS2Merged/properties'
+hips_survey = HipsSurveyProperties.fetch(url)
+geometry = WCSGeometry.create_simple(
+    skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
+    width=2000, height=1000, fov="3 deg",
+    coordsys='galactic', projection='AIT'
+)
+data = make_sky_image(geometry=geometry, hips_survey=hips_survey, tile_format='jpg')
```
cdeil: I would suggest to just add the "fits" plotting example that is working here for now.
JPEG drawing isn't working yet, and is probably best for a separate PR, no?

hips/draw/tests/test_simple.py
```diff
@@ -38,7 +38,7 @@ def get_test_tiles(file_format, survey, order, ipix_list):
 @remote_data
 @requires_hips_extra()
 @pytest.mark.parametrize('pars', draw_sky_image_pars)
-def test_draw_sky_image(pars):
+def test_draw_sky_image(pars, tmpdir):
```
cdeil: Accidental change? Remove?

(just passing `tmpdir` has no effect, you would have to use it from within the test function.

cdeil:
I think I found a way to prevent `docs/plot.py` from being collected (and failing) as a doctest.
Add the line `doctest_norecursedirs = docs/plot.py` to `setup.cfg` at the end of the `tool:pytest` section.

My other suggestion would be that you plot the sky image with `wcsaxes`, so that it has coordinate axes.

So I think if you limit this PR to getting this image into the high-level docs, this PR could be finalised quickly.

adl1995:
@cdeil I made the changes. Let's see if the test case passes now...

cdeil:
What do you get locally? Please share a screenshot here.

cdeil:
Works for me locally, I get this:
image

adl1995:
Here it is:
plot

docs/getting_started.rst
```diff
@@ -54,7 +54,9 @@ Now you can then save the sky image to local disk e.g. FITS file format::
 
 or plot and analyse the sky image however you like.
 
-TODO: show how to plot the image with `astropy.visualization.wcsaxes`.
+The following plot was created using `astropy.visualization.wcsaxes`.
+
+.. plot:: plot.py
```
bsipocz: try to add ``.. doctest-skip::`` here

docs/plot.py
```diff
@@ -0,0 +1,19 @@
...
+hips_survey = HipsSurveyProperties.fetch(url)
+geometry = WCSGeometry.create_simple(
+    skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
+    width=2000, height=1000, fov="3 deg",
+    coordsys='galactic', projection='AIT'
+)
+data = make_sky_image(geometry=geometry, hips_survey=hips_survey, tile_format='fits')
+ax = plt.subplot(projection=geometry.wcs)
+ax.imshow(data, origin='lower', cmap='gray')
+ax.grid(color='white', ls='solid')
```
cdeil: Suggest to remove the `ax.grid` call here, and to change the image scale to sqrt or something so that it's not so dark (at the moment, it looks almost all black).

See http://astropy.readthedocs.io/en/latest/api/astropy.visualization.mpl_normalize.simple_norm.html
I don't know if you can pass `norm='sqrt'` to `plt.imshow` directly, or if you have to do:
```
from astropy.visualization.mpl_normalize import simple_norm
norm = simple_norm(data, 'sqrt')
```
and then pass this norm.

Also: we could make the image a little bit smaller (e.g. 1600, 900) or something like that) so that the docs build and tests run a bit faster, but look just as nice / test just as much. If you want, you could do that change in this PR, or I or you will do it tomorrow.

docs/getting_started.rst
```diff
@@ -54,7 +54,9 @@ Now you can then save the sky image to local disk e.g. FITS file format::
 
 or plot and analyse the sky image however you like.
 
-TODO: show how to plot the image with `astropy.visualization.wcsaxes`.
+The following plot was created using `astropy.visualization.wcsaxes`.
```
cdeil: Remove `.` at the end of the line and improve the text a bit, e.g. "If you execute the example above, you will get this sky image", i.e. make it clear how "this plot" relates to the example given in this secion.

cdeil:
This is looking good, this can go in soon.

I left two inline commments. My last comment would be to maybe change the name of the script a bit from `plot.py` to something more specific, at least `plot_fits.py`.
Note that already tomorrow, we'll get a second script that plots an RGB image, and then a few more scripts in the docs in the future still.

adl1995:
@cdeil I think this is ready to be merged now.

cdeil:
I worked, image is here: https://hips.readthedocs.io/en/latest/getting_started.html

🎈 🎈 🎈

cdeil:
Just for future reference, in case we discover some bug or change that example, the image at the moment looks like this:

PR #70, opened 2017-07-14
Misc changes / cleanup before v0.1

cdeil:
This pull request contains some misc changes and cleanup I'd like to make before releasing v0.1:

- [x] Don't expose `hips.utils` in the public docs. Make it simple for end-users and expose everything they should know about in the top-level `hips` namespace. The one thing from `hips.utils` that they need is `WCSGeometry`, which I import into the top-level `hips` namespace.
- [x] Shorten the docs landing page, move the content to a new "about" page.
- [x] Update drawing description to what we do in the code (e.g. projective transform, not affine transform with triangles)
- [x] Update getting started guide to include a color image draw example.
- [x] Misc code, test and docstring cleanup

@tboch - Do you want to review this today or are you already on vacation?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,9 @@
-0.1 (unreleased)
-----------------
+0.1
+---
 
-- Set up initial Python package [#1]
+This first version of the hips package was released on July 15, 2017.
+It contains a first implementation to fetch and draw tiles.
+
+The ``hips`` package started as a project developed as part of
+Google summer of code 2017, i.e. planning in early 2017 and
+coding in June 2017.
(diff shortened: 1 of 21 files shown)


cdeil:
@tboch @adl1995 - I've added an "about" page, and updated the "drawing algo" page and done other docs changes. If you have time to review them now, please do. Otherwise I'll merge after lunch and then you can still review and comment in the coming days of course!

cdeil:
I've added an example drawing `jpg` from Fermi color to the high-level docs.

The result is bad:

image

I've never seen a successful RGB draw, and I don't know yet what the issue is.
@adl1995 - Have you made any progress on that yet?

If no, I would like to merge this PR after lunch, and then work on this issue in the afternoon.

tboch:
@cdeil sorry, I'm already off

PR #73, opened 2017-07-17
Fix RGB dtype issue

adl1995:
@cdeil As discussed on Slack, I have updated the `dtype` so that RGB drawing works fine.

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -63,6 +63,11 @@ class SimpleTilePainter:
         self._tiles = None
 
     @property
+    def image(self) -> np.ndarray:
+        """All sky image converted to HiPS tile format."""
+        return self.float_image.astype(self.tiles[0].data.dtype)
+
+    @property
     def draw_hips_order(self) -> int:
         """Compute HiPS tile order matching a given image pixel size."""
         # Sky image angular resolution (pixel size in degrees)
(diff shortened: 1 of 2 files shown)


hips/draw/simple.py
```diff
@@ -63,6 +63,10 @@ def __init__(self, geometry: WCSGeometry, hips_survey: HipsSurveyProperties, til
         self._tiles = None
 
     @property
+    def image(self):
```
cdeil: Always add a one-line docstring mentioning what it is, what object type is returned, and also add a type annotation.

adl1995:
I just updated the docstring. Here is the output image I'm getting:

jpg_out

cdeil:
@adl1995 - You must have been looking at an outdated image on the docs page.
It's all black now: https://hips.readthedocs.io/en/latest//plot_fits.png
I'll make a commit in master shortly to fix.

cdeil:
Follow-up commits in master: 9b52a7a and 60a5d95

cdeil:
I had to wipe RTD and trigger a completely fresh build, for the `plot` scripts from the docs build to re-run.
It worked https://readthedocs.org/projects/hips/builds/5698160/ and now the images on
http://hips.readthedocs.io/en/latest/getting_started.html are correct, except for the last remaining drawing issue #74 .

PR #75, opened 2017-07-18
Add methods for drawing markers over sky image

adl1995:
As outlined in this Jupyter notebook (https://github.com/adl1995/HIPS-to-Py/blob/master/examples/JPG_tile_draw_issue.ipynb), I have added two methods for drawing markers and lines over an all-sky image.

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -1,7 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """HiPS tile drawing -- simple method."""
-from typing import Tuple, List
 import numpy as np
+from typing import Tuple, List
 from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
 from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta
@@ -9,8 +9,7 @@ from ..utils import WCSGeometry, compute_healpix_pixel_indices, get_hips_order_f
 
 __all__ = [
     'make_sky_image',
-    'SimpleTilePainter',
-
+    'SimpleTilePainter'
 ]
 
 __doctest_skip__ = [
@@ -178,9 +177,47 @@ class SimpleTilePainter:
 
         return image
 
+    def draw_hips_tile_grid(self) -> None:
+        """Draw lines on the output image (mainly used for debugging)."""
+        import matplotlib.pyplot as plt
+        for tile in self.tiles:
+            corners = tile.meta.skycoord_corners.transform_to(self.geometry.celestial_frame)
+            ax = plt.subplot(projection=self.geometry.wcs)
+            ax.plot(corners.data.lon.deg, corners.data.lat.deg,
+                    'red', lw=1, transform=ax.get_transform('icrs'))
+        ax.imshow(self.image, origin='lower')
+
     def run(self) -> None:
         """Run all steps of the naive algorithm."""
         self.float_image = self.draw_tiles()
+
+
+def draw_debug_image(geometry: WCSGeometry, tile: HipsTile, image: np.ndarray) -> None:
+    """Draw markers on the output image (mainly used for debugging).
+
+    The following denotes their correspondence:
+    * red <=> North
+    * green <=> West
+    * blue <=> South
+    * yellow <=> East
+
+    Parameters
+    ----------
+    geometry : `~hips.utils.WCSGeometry`
+        Geometry of the output image
+    tile : HipsTile
+        HiPS tile
+    image : np.ndarray
+        Image containing HiPS tiles
+    """
+    import matplotlib.pyplot as plt
+    corners = tile.meta.skycoord_corners.transform_to(geometry.celestial_frame)
+    colors = ['red', 'green', 'blue', 'yellow']
+    ax = plt.subplot(projection=geometry.wcs)
+    for index, corner in enumerate(corners):
+        ax.scatter(corner.data.lon.deg, corner.data.lat.deg,
+                   s=80, transform=ax.get_transform('icrs'), color=colors[index])
+    ax.imshow(image, origin='lower')
 
 
 def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile_format: str) -> np.ndarray:
--- a/hips/draw/tests/test_simple.py
+++ b/hips/draw/tests/test_simple.py
@@ -5,7 +5,7 @@ from numpy.testing import assert_allclose
 from astropy.coordinates import SkyCoord
 from astropy.tests.helper import remote_data
 from ...tiles import HipsSurveyProperties
-from ..simple import make_sky_image, SimpleTilePainter
+from ..simple import make_sky_image, SimpleTilePainter, draw_debug_image
 from ...utils.wcs import WCSGeometry
 from ...utils.testing import make_test_wcs_geometry, requires_hips_extra
 
@@ -95,3 +95,9 @@ class TestSimpleTilePainter:
     def test_run(self):
         self.simple_tile_painter.run()
         assert_allclose(self.simple_tile_painter.image[200, 994], 2120)
+
+    def test_draw_hips_tile_grid(self):
+        self.simple_tile_painter.draw_hips_tile_grid()
+
+    def test_draw_debug_image(self):
+        draw_debug_image(self.geometry, self.simple_tile_painter.tiles[3], self.simple_tile_painter.image)


hips/draw/simple.py
```diff
@@ -1,7 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """HiPS tile drawing -- simple method."""
-from typing import Tuple, List
 import numpy as np
+from typing import Tuple, List
+import matplotlib.pyplot as plt
```
cdeil: MPL is an optional dependency for the `hips` package. One should be able to draw and e.g. save a sky image without having it installed. If you put the MPL import here at the top of `simple.py`, that's not possible.

So instead, for optional dependencies, what we do is to put a "delayed" import on the first line of each function or method where you use it. That way, the import only happens if a user uses some functionality that requires the optional dependency.

hips/draw/simple.py
```diff
@@ -178,6 +179,40 @@ def draw_tiles(self) -> np.ndarray:
 
         return image
 
+    def draw_markers_tiles(self, tile_index: int) -> None:
```
cdeil: Isn't `tile_index` a bit confusing, because it's the index in the tiles list, not the HEALPix index of the tile?
Also, the implementation with having everything in a try except clause is not good.

My suggestion would be that you make this a standalone function, and you pass in the `tile` and `image`.

hips/draw/simple.py
```diff
@@ -178,6 +179,40 @@ def draw_tiles(self) -> np.ndarray:
...
+        ----------
+        tile_index : int
+            HiPS tile index
+        """
+        try:
+            radec = self.tiles[tile_index].meta.skycoord_corners
+            colors = ['red', 'green', 'blue', 'yellow']
+            ax = plt.subplot(projection=self.geometry.wcs)
+            for i in range(len(radec.data.lon.deg)):
+                ax.scatter(radec.data.lon.deg[i], radec.data.lat.deg[i], s=80, transform=ax.get_transform('icrs'), color=colors[i])
```
cdeil: Break this line, it's too long.

hips/draw/simple.py
```diff
@@ -178,6 +179,40 @@ def draw_tiles(self) -> np.ndarray:
...
+            raise IndexError(f'Index: {tile_index} is out of bounds')
+
+    def draw_lines_tiles(self) -> None:
+        """Draw lines on the output image."""
+        for tile in self.tiles:
+            radec = tile.meta.skycoord_corners
+            ax = plt.subplot(projection=self.geometry.wcs)
+            ax.plot(radec.data.lon.deg, radec.data.lat.deg, 'red', lw=1, transform=ax.get_transform('icrs'))
+        ax.imshow(self.image, origin='lower')
+        plt.show()
```
cdeil: You should remove the `plt.show` call here and instead let users do that call if needed.
This will prevent the hanging of the tests at https://travis-ci.org/hipspy/hips/jobs/254765525#L1215 because a GUI window pops up, and it will also prevent a "double plotting" bug in Jupyter notebooks, where if people are using the common `%matplotlib inline`, the Figure will be plotted twice if you call `plt.show` here.

hips/draw/simple.py
```diff
@@ -178,6 +179,40 @@ def draw_tiles(self) -> np.ndarray:
...
+            colors = ['red', 'green', 'blue', 'yellow']
+            ax = plt.subplot(projection=self.geometry.wcs)
+            for i in range(len(radec.data.lon.deg)):
+                ax.scatter(radec.data.lon.deg[i], radec.data.lat.deg[i], s=80, transform=ax.get_transform('icrs'), color=colors[i])
+            ax.imshow(self.image, origin='lower')
+            plt.show()
+        except IndexError:
+            raise IndexError(f'Index: {tile_index} is out of bounds')
+
+    def draw_lines_tiles(self) -> None:
```
cdeil: Suggest to rename to `draw_hips_tile_grid`.

adl1995:
@cdeil Just updated. Please check.

hips/draw/simple.py
```diff
@@ -10,6 +10,7 @@
 __all__ = [
     'make_sky_image',
     'SimpleTilePainter',
+    'draw_markers_tiles'
```
cdeil: I would suggest to not expose this to end-users, i.e. not list it here.
It's a "hidden feature" that only experts like you and me use.

There's always a tradeoff, what to expose to end-users. Of course it's nice to have everything documented, but then again if the API becomes too large, users will get lost. So it's a judgement call for every function, and in this case, I think we don't need to expose this to end-users.

hips/draw/simple.py
```diff
@@ -178,11 +179,48 @@ def draw_tiles(self) -> np.ndarray:
 
         return image
 
+    def draw_hips_tile_grid(self) -> None:
```
cdeil: Contrary to what I said before, this isn't a good name for this method.
It doesn't just draw the tile grid, it also draws the sky iamge.

So my suggestion would be to rename to "draw_debug_image", and maybe even to split out the two parts into `_draw_sky_image` and `_draw_hips_tile_grid` functions and call the two from `draw_debug_image`?

hips/draw/simple.py
adl1995: @cdeil Okay. Just removed it.

hips/draw/simple.py
adl1995: @cdeil Then `_draw_sky_image` function would only contain the line `ax.imshow(image, origin='lower')`?

hips/draw/simple.py
```diff
@@ -178,11 +179,48 @@ def draw_tiles(self) -> np.ndarray:
...
+        """Draw lines on the output image (mainly used for debugging)."""
+        import matplotlib.pyplot as plt
+        for tile in self.tiles:
+            radec = tile.meta.skycoord_corners
```
cdeil: I don't think this will work if you have tiles in the "galactic" frame.
In that case, `tile.meta.skycoord_corners` will be a skycoord in the Galactic frame, i.e. not `radec` like you call the local variable.

There's several ways to handle this, to make it work for any combination of frames.
I would suggest you do
```
pos = tile.meta.skycoord_corners.transform_to(self.geometry.celestial_frame)
```
here, and then below pass `transform=ax.get_transform('world')`

Also, please make a notebook somewhere (not in this repo) and link to it, showing that this code works for two examples: galactic hips tiles drawn on galactic image, icrs hips tiles drawn on galactic image.

hips/draw/simple.py
cdeil: I guess. Probably it's not needed then to make it modular and separate?
Maybe just rename this one function and give it a better name and say in the docstring that it does these two things?

@adl1995 - your choice how to finalise this!

hips/draw/simple.py
adl1995: Is it ok if I access the values like this: `pos.data.lon.deg[i]`?

cdeil: Yes. That's how I would do it also. This should give you just the lon / lat values in the given frame of the `pos` Skycoord.

hips/draw/simple.py
```diff
@@ -178,11 +177,52 @@ def draw_tiles(self) -> np.ndarray:
...
+    tile : HipsTile
+        HiPS tile
+    image : np.ndarray
+        Image containing HiPS tiles
+    """
+    import matplotlib.pyplot as plt
+    pos = tile.meta.skycoord_corners.transform_to(geometry.celestial_frame)
+    colors = ['red', 'green', 'blue', 'yellow']
+    ax = plt.subplot(projection=geometry.wcs)
+    for i in range(len(pos.data.lon.deg)):
```
cdeil: I think you can just do this, no?
```
corners = tile.meta.skycoord_corners.transform_to(geometry.celestial_frame)
for corner in corners:
    # corner is a skycoord here, i.e. `corner.data.lon.deg` and `corner.data.lat.deg` will give you the right values
```

hips/draw/simple.py
```diff
@@ -178,11 +177,52 @@ def draw_tiles(self) -> np.ndarray:
...
+        """Draw lines on the output image (mainly used for debugging).
+
+        A Jupyter notebook showcasing this functionality can be found
+        here: https://github.com/adl1995/HIPS-to-Py/blob/master/examples/Draw_markers_plots_image.ipynb
+        """
+        import matplotlib.pyplot as plt
+        for tile in self.tiles:
+            radec = tile.meta.skycoord_corners
+            ax = plt.subplot(projection=self.geometry.wcs)
+            ax.plot(radec.data.lon.deg, radec.data.lat.deg, 'red', lw=1, transform=ax.get_transform('icrs'))
```
cdeil: Please break the line.

hips/draw/simple.py
```diff
@@ -178,11 +177,52 @@ def draw_tiles(self) -> np.ndarray:
 
         return image
 
+    def draw_hips_tile_grid(self) -> None:
+        """Draw lines on the output image (mainly used for debugging).
+
+        A Jupyter notebook showcasing this functionality can be found
+        here: https://github.com/adl1995/HIPS-to-Py/blob/master/examples/Draw_markers_plots_image.ipynb
```
cdeil: Don't link to your notebook from the docstring, just from a Github comment.
External code using `hips` will break very quickly, as we refactor and improve our `hips` package.

hips/draw/tests/test_simple.py
```diff
@@ -40,17 +40,17 @@
...
+# @pytest.mark.parametrize('pars', make_sky_image_pars)
+# def test_make_sky_image(pars):
+#     hips_survey = HipsSurveyProperties.fetch(url=pars['url'])
+#     geometry = make_test_wcs_geometry(case=2)
+#     image = make_sky_image(geometry=geometry, hips_survey=hips_survey, tile_format=pars['file_format'])
+#     assert image.shape == pars['shape']
+#     assert image.dtype == pars['dtype']
+#     assert_allclose(np.sum(image), pars['data_sum'])
+#     assert_allclose(image[200, 994], pars['data_1'])
+#     assert_allclose(image[200, 995], pars['data_2'])
```
cdeil: Remove this change where you commented out code.
If you want to just run certain tests, use `pytest` to select which ones.

adl1995:
@cdeil I made the update, but now the grid lines from the second image are gone as well.

cdeil:
@adl1995 - I think I know what's wrong. I'll merge now and fix in master and link to the fixup commit from here.

cdeil:
I think with 524bb43 it should work properly. @adl1995 - Please try again with latest master.

PR #77, opened 2017-07-18
Fix tile_width property in HipsSurveyProperties class

adl1995:
I encountered this issue when playing with certain surveys that are missing the `hips_tile_width` property key, such as this one (http://alasky.u-strasbg.fr/PLANCK/R2/HFI_Color_353_545_857/properties). They simply failed when I tried to access the property, this PR will fix this issue.


cdeil:
OK, thanks!

Issue #74, opened 2017-07-17
Incorrect corner order of HiPS during stitching


adl1995:
This issue relates with incorrect order for HiPS tiles, this leads to incorrect stitching of tiles. The notebook related with this can be found here (https://github.com/adl1995/HIPS-to-Py/blob/master/examples/JPG_tile_draw_issue.ipynb).

cdeil:
Related issue with infos from Thomas: #18

cdeil:
I think I found the origin of the tile corner order issue. with JPG and PNG, the tile images are flipped in the y direction. See Figure 6 in http://www.ivoa.net/documents/HiPS/20160623/WD-HiPS-1.0-20160623.pdf

I'll make a PR today.

cdeil:
Fixed in #78

Issue #18, opened 2017-06-03
How to find which corner is which in tile drawing


cdeil:
@adl1995 and I started to play with `healpy.boundaries` and try to prototype the affine transformation drawing method (see end of this notebook):
https://github.com/adl1995/HIPS-to-Py/blob/master/examples/draw-healpix-stuff.ipynb

One issue we ran into is https://github.com/healpy/healpy/issues/393 , the other (somewhat related) blocker for implementing a first working version is that we don't know how to find out which of the four corners returned by `healpy.boundaries` corresponds to which corner of the HIPS tile. I don't see anything in `healpy` that would let us compute this directly.

@tboch - Could you clarify this point in the description?
https://hips.readthedocs.io/en/latest/drawing_algo.html#naive-algorithm
(or directly here in a comment if you prefer).

tboch:
Figure 4, page 10 of HiPS document http://ivoa.net/documents/HiPS/20170406/PR-HIPS-1.0-20170406.pdf should help
I'll update the drawing algo document to detail this.

tboch:
Looking closely after this, I found out that healpy.boundaries always returns the corners in this order: N, W, S, E where N (resp. W, S, E) is the corner roughly pointing towards the North (resp. West, South and East) : 
![image](https://user-images.githubusercontent.com/5598168/26928101-fef34778-4c54-11e7-90fc-1ae81b3d9592.png)

The corresponding tile corners are as follow:
![image](https://user-images.githubusercontent.com/5598168/26928164-4f26b446-4c55-11e7-9a1f-f0efa6ad0e31.png)

cdeil:
@tboch - Thanks!

We should make sure this info makes it into our docs before closing this issue.

@adl1995 - one thing you could do now is make a PR to improve the docstring of https://hips.readthedocs.io/en/latest/api/hips.utils.boundaries.html to mention the order of corners returned.
Actually the docstring currently only says
> Returns an array containing the angle (theta and phi) in radians.
and doesn't mention "corners" at all -> please change this.
Another change I would suggest is to go from default `nest=False` to `nest=True`, because this is just a helper function we use in hips, and HiPS always uses the nested  scheme.
[...]

PR #78, opened 2017-07-19
Fix tile corner order for PNG and JPEG

cdeil:
This PR fixes the tile corner order in drawing for PNG and JPEG tiles.

Mentioned in #74 and also #18 .

I make a few images and they all looked OK now.

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -91,22 +91,14 @@ class SimpleTilePainter:
             healpix_frame=self.hips_survey.astropy_frame,
         )
 
-    @property
-    def dst(self) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


cdeil:
I'm merging this now.

@adl1995 - I think your "draw HiPS tile grid" functionality still doesn't work for PNG and JPEG tiles, because it would need the same fix? Probably the implementation should be changed to plot lines in pixel coordinates, and to re-use this functionality to compute which corner is which from NWES ?
Please update the notebook and file an issue with the result and link to the notebook if still not fully OK. And then we can discuss how best to change the implementation to get correct HiPS tile rectangles drawn in all cases.

Issue #83, opened 2017-07-19
Introduce method for creating children tiles


adl1995:
As discussed during telecon today with @cdeil (also mentioned in the [precise algorithm](https://hips.readthedocs.io/en/latest/drawing_algo.html#tile-distortion-issue)), I will open a PR adding a method named: `make_child_tiles`(?) which will take as parameter the father tile. It will then split this into four children tiles by slicing the father tile into four quarters. It will use something like: tile[0, width / 4] and then call [HipsTile.from_numpy](http://hips.readthedocs.io/en/latest/api/hips.HipsTile.html?highlight=from_numpy#hips.HipsTile.from_numpy) class method. The new tiles will be labeled according to the scheme mentioned in the [HiPS paper](http://www.ivoa.net/documents/HiPS/20160623/WD-HiPS-1.0-20160623.pdf), page 31, titled "Children of the tile".

cdeil:
I'm never sure with naming at first and then find that only over time as one starts to call the functionality good names and locations for code are found. In this case, I think I would start by putting it as a property on `HipsTile` called `children` which returns the list of 4 child tiles.

There should be a good docstring and a parameterised test using these tiles we already have with the different formats: https://github.com/hipspy/hips/blob/master/hips/tiles/tests/test_tile.py#L54

Otherwise, yes, your description sounds good to me ... go ahead!

adl1995:
Done in #85. Closing this issue now.

PR #85, opened 2017-07-20
Add property 'children' for creating four children tiles from parent tile

adl1995:
As outlined in #83, I have added a property named `children` in `HipsTile` class, this splits the parent tile into four children tiles. I also added a some test cases for the first child.

### DIFF
--- a/hips/tiles/tests/test_tile.py
+++ b/hips/tiles/tests/test_tile.py
@@ -59,8 +59,13 @@ HIPS_TILE_TEST_CASES = [
 
         dtype='int16',
         shape=(512, 512),
+        child_order=4,
+        child_shape=(256, 256),
         pix_idx=[[510], [5]],
         pix_val=[3047],
+        child_ipix=[1852, 1853, 1854, 1855],
+        child_pix_idx=[[0], [255]],
+        child_pix_val=[2407, 2321, 2465, 2835],
     ),
     dict(
         meta=dict(order=3, ipix=463, file_format='jpg'),
@@ -69,8 +74,13 @@ HIPS_TILE_TEST_CASES = [
 
         dtype='uint8',
         shape=(512, 512, 3),
+        child_order=4,
+        child_shape=(256, 256, 3),
         pix_idx=[[510], [5]],
         pix_val=[[116, 81, 61]],
+        child_ipix=[1852, 1853, 1854, 1855],
+        child_pix_idx=[[0], [255]],
+        child_pix_val=[[[255, 241, 225]], [[109, 95, 86]], [[245, 214, 211]], [[137, 97, 87]]],
     ),
     dict(
         meta=dict(order=6, ipix=6112, file_format='png'),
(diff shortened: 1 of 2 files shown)


cdeil:
The splitting isn't correct. For a tile of shape (512, 512), all the children should have shape (256, 256). You need to fix the implementation.

For the tests: please make a separate `test_children` method, don't mix the tests for the `children` property in with `test_read`.

Finally, even if the output shape of the children is correct, I guess it'll still not be clear whether their orientation is correct, or whether they are flipped in x or y or rotated. You could make a notebook drawing once the parent tile, and then the children, and show that the resulting sky image is (almost, apart from small distortiion correction) the same. I leave it up to you if you want to do this check now, or in the next PR where you use these children for drawing.

adl1995:
@cdeil I have updated the implementation. Please check if it's correct now. For the Jupyter notebook, I would prefer to create it along with the next pull request.

cdeil:
The implementation now splits into 4 parts with the correct shapes. But I still don't know if it's correct. You are putting the child with `data[0: w, 0: w]` to correspond to `meta.ipix * 4 + 0`, but that is just a guess, right? The correspondence which data sub-part goes with which HEALPIX ipix for the child is unclear and non-obvious (and thus probably incorrect).

My preference would be that you debug this here, because the next pull request will be even more complex, with changes to the painter, so debugging this utils function there in addition could take very long. Whereas this here is a very small pull request, and you will manage to understand, debug, fix and then document via a comment in the code the child tile order and orientation.

It's not so hard, it should be ~ 20 lines of code and take an hour: you just make a notebook where you draw a single tile, and then draw the four sub-tiles. Either the output images are correct, or you'll see that the children are in the wrong place or flipped and then adjust your code until it's correct.

--

For the test: remove all changes to the `test_read()` method. And change `test_children` to assert something about each of the four child tiles that establishes the current behaviour. Concretely I would assert on the `ipix` of each child, and the pixel value at pixel index `(0, 255)`.

hips/tiles/tests/test_tile.py
```diff
@@ -112,12 +124,26 @@ def test_read(self, pars):
...
+        assert_equal(data[pars['pix_idx']], pars['pix_val'])
+
+        assert child1_data.shape == pars['child1_shape']
+        assert_equal(child1_data[pars['child1_pix_idx']], pars['child1_pix_val'])
+
+    @requires_hips_extra()
+    @pytest.mark.parametrize('pars', HIPS_TILE_TEST_CASES)
+    def test_children(self, pars):
+        tile = self._read_tile(pars)
+        child1_data = tile.children[0].data
```
cdeil: Here I would suggest to put this on a separate line:
```
child_tiles = tile.children
```
Then you can on the following lines access meta and data from each child as you like, e.g. you could do:
```
ipix = [_.meta.ipix for _ in child_tiles]
```
and then assert on this `ipix` list, to establish the current behaviour of the function about the order of child tiles returned.

adl1995:
@cdeil I just tried displaying the found children tiles and compared it with the father tile, the result looks good to me. Link to notebook: https://github.com/adl1995/HIPS-to-Py/blob/master/examples/Tile%20distortion%20issue%20-%20Fix.ipynb

cdeil:
Now project these tiles onto the sky image, and see if they land in the right place / with the right orientation. If it's not clear to you what to do here, let's chat on Slack.

adl1995:
@cdeil I updated the code according to your suggestions. The drawing code isn't updated yet, I will do that in the next PR.

cdeil:
Mui bien!

PR #87, opened 2017-07-23
Remove print statement from test case

adl1995:
I found this `print` statement in `test_allsky.py` and thought best to remove it.


cdeil:
@adl1995 - Thanks!

I'm working on those tests now (adding a FITS case and choosing better tiles / pixels for the asserts) and have removed the print statement as part of that PR which I'll make later today.

Issue #84, opened 2017-07-19
Introduce precise drawing algorithm using a boolean


adl1995:
The `SimpleTilePainter` class will take as a parameter `reduce_deformations`(?) boolean which will be used before the actual drawing begins. If this is set to `True`, it will traverse the `tiles` list and check if it falls under the criteria mentioned in [Tile distortion issue](https://hips.readthedocs.io/en/latest/drawing_algo.html#tile-distortion-issue), if it does, it will split this father tile into four children tiles and will extend the current list to contain the new labels.

Drawing procedure will remain exactly the same.

cdeil:
Actually, I noticed just now that we might need some larger changes to how the `SimpleTilePainter` works. The thing is that the criteria whether to split a tile in kids is based on the projected tile pixel corner coordinates. (Is that right? What do you think from the description?)

At the moment we compute the tile list and then only compute the tile corners in the output image for the tile transform later when drawing. Maybe it would be better to change the implementation to compute the tile corner positions in the output image at the start, so that it's available go compute this second tile list? So overall it would be more like:

1. compute the full plan of how drawing will work, i.e. the complete tile and child tile lists and their corner coordinates
2. execute the plan, i.e. fetch the tiles and draw them

I would suggest you do #83 first and when that's done either make a PR here with how you'd implement it, or we discuss some more before you start coding. Let me know if you  have any questions.

adl1995:
@cdeil So, basically this means to compute the corners at the beginning of `draw_tiles` function [here](https://github.com/hipspy/hips/blob/master/hips/draw/simple.py#L162) and then pass the computed corners to `warp_image`function [here](https://github.com/hipspy/hips/blob/master/hips/draw/simple.py#L165), which will then pass it to `self.projection` [here](https://github.com/hipspy/hips/blob/master/hips/draw/simple.py#L154)?

Also, these corners will be modified (extended) when a certain tile passes a condition?

cdeil:
@adl1995 - I think this is the most complicated thing you've done so far in the `hips` package, and I fear that if you try to do it in one PR it will completely change the code organisation in `draw.py` and take a week and it will block me from touching the code in `draw.py` to do #79 . So I'm wondering if it's possible to go step by step here, and you and I make small pull requests that change / improve or add something to the existing code (each 1 day of coding / code review max), without touching many of the methods.

My preference / suggestion here would be that you start to implement this part as a first PR:
[...]

PR #89, opened 2017-07-24
Add method to_split in SimpleTilePainter for checking whether to split a tile

adl1995:
As mentioned in issue #84, I have added two methods `to_split` and `_measure_tile_shape` which are used for checking whether a tile needs to be split or not. The next step is to perform the actual splitting and modify the projection part to take in the correct `tile_width` (half, in case a tile is split).

### DIFF
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -172,6 +172,29 @@ class SimpleTilePainter:
         ax.imshow(self.image, origin='lower')
 
 
+def _measure_tile_shape(corners: tuple) -> List[list]:
+    """Compute length of tile edges and diagonals."""
+    x, y = corners
+
+    def compute_distance(i: int, j: int) -> float:
+        """Compute distance between two points."""
+        return np.sqrt((x[i] - x[j]) ** 2 + (y[i] - y[j]) ** 2)
+
+    edges = [compute_distance((i + 1) % 4, i) for i in range(4)]
+    diagonals = [compute_distance(0, 2), compute_distance(1, 3)]
+    ratio = float(np.min(diagonals) / np.max(diagonals))
+
+    return [edges, diagonals, ratio]
+
+def _is_tile_distorted(corners: tuple) -> bool:
+    """Implement tile splitting criteria as mentioned in :ref:`drawing_algo` page."""
+    edges, diagonals, ratio = _measure_tile_shape(corners)
+
+    return max(edges) > 300 or \
+           max(diagonals) > 150 or \
+           ratio < 0.7
+
+
 def tile_corner_pixel_coordinates(width, file_format) -> np.ndarray:
     """Tile corner pixel coordinates for projective transform.
 
--- a/hips/draw/tests/test_simple.py
+++ b/hips/draw/tests/test_simple.py
@@ -5,7 +5,7 @@ from numpy.testing import assert_allclose
 from astropy.coordinates import SkyCoord
 from astropy.tests.helper import remote_data
 from ...tiles import HipsSurveyProperties
-from ..simple import make_sky_image, SimpleTilePainter, plot_mpl_single_tile
+from ..simple import make_sky_image, SimpleTilePainter, plot_mpl_single_tile, _is_tile_distorted, _measure_tile_shape
 from ...utils.wcs import WCSGeometry
 from ...utils.testing import make_test_wcs_geometry, requires_hips_extra
 
@@ -86,6 +86,7 @@ class TestSimpleTilePainter:
             width=2000, height=1000, fov=pars['fov'],
             coordsys='icrs', projection='AIT',
         )
+
         simple_tile_painter = SimpleTilePainter(geometry, self.hips_survey, 'fits')
         assert simple_tile_painter.draw_hips_order == pars['order']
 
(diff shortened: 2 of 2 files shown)


adl1995:
@cdeil I just made the updates.

hips/draw/simple.py
```diff
@@ -117,6 +117,38 @@ def _fetch_tiles(self) -> HipsTile:
...
+
+        diagonals = []
+        diagonals.append(compute_distance([x[2], x[0]], [y[2], y[0]]))
+        diagonals.append(compute_distance([x[3], x[1]], [y[3], y[1]]))
+
+        ratio = float(np.min(diagonals) / np.max(diagonals))
+
+        return [edges, diagonals, ratio]
+
+    def is_tile_distorted(self, corners: tuple) -> bool:
```
cdeil: Make this a @staticmethod

hips/draw/tests/test_simple.py
```diff
@@ -101,3 +102,21 @@ def test_draw_debug_image(self):
         tile = self.painter.tiles[3]
         image = self.painter.image
         plot_mpl_single_tile(self.geometry, tile, image)
+
+    def test_to_split(self):
```
cdeil: Rename to `test_ is_tile_distorted`.

hips/draw/simple.py
```diff
@@ -117,6 +117,38 @@ def _fetch_tiles(self) -> HipsTile:
             tile = HipsTile.fetch(tile_meta, url)
             yield tile
 
+    @staticmethod
+    def _measure_tile_shape(corners: tuple) -> List[list]:
+        """Compute tile edges and diagonals."""
+
+        def compute_distance(i: list, j: list) -> float:
```
cdeil: I find this way you've implemented `compute_distance` hard to read.
My suggestion would be that you move the
```
x, y = corners
```
to the top and then do this:
```
def compute_distance(i, j):
    return np.sqrt( (x[i] - x[j]) ** 2 + (y[i] - y[j]) ** 2)
```
i.e. have i and j as the integer index of the corner.

Then you should be able to call it like this:
```
edge_lengths = [compute_distance(i, (i+1) %4) for i in range(4)]
```
and like this
```
diagonal_lengths = [compute_distance(0, 2), compute_distance(1, 3)]
```
Note that I didn't try this code, I just typed it here.
Can you please try to change the implementation to work this way?

hips/draw/simple.py
```diff
@@ -117,6 +117,38 @@ def _fetch_tiles(self) -> HipsTile:
...
+
+        ratio = float(np.min(diagonals) / np.max(diagonals))
+
+        return [edges, diagonals, ratio]
+
+    def is_tile_distorted(self, corners: tuple) -> bool:
+        """Implement tile splitting criteria as mentioned in :ref:`drawing_algo` page."""
+        edges, diagonals, ratio = self._measure_tile_shape(corners)
+
+        return any(i > 300 for i in edges) or \
```
cdeil: I think I would find `max(edge_lengths) > 300` and equivalently for diagonals easier to read.

adl1995:
@cdeil Okay, I made the changes. However, the test results vary from before. I'm not sure if these are the correct ones or if the previous ones were.

hips/draw/simple.py
```diff
@@ -117,6 +117,30 @@ def _fetch_tiles(self) -> HipsTile:
...
+    @staticmethod
+    def _measure_tile_shape(corners: tuple) -> List[list]:
+        """Compute tile edges and diagonals."""
+        x, y = corners
+
+        def compute_distance(i: int, j: int) -> float:
+            """Compute distance between two points."""
+            return np.sqrt((x[i] - x[j]) ** 2 + (y[i] - y[j]) ** 2)
+
+        edges = [compute_distance((i + 1) % 3, i) for i in range(4)]
```
cdeil: You now have `%3`. Shouldn't this be `%4` instead?

hips/draw/simple.py
```diff
@@ -117,6 +117,30 @@ def _fetch_tiles(self) -> HipsTile:
             tile = HipsTile.fetch(tile_meta, url)
             yield tile
 
+    @staticmethod
+    def _measure_tile_shape(corners: tuple) -> List[list]:
+        """Compute tile edges and diagonals."""
```
cdeil: "Compute tile edges and diagonals" -> "Compute length of tile edges and diagonals"

hips/draw/simple.py
```diff
@@ -117,6 +117,30 @@ def _fetch_tiles(self) -> HipsTile:
...
+            return np.sqrt((x[i] - x[j]) ** 2 + (y[i] - y[j]) ** 2)
+
+        edges = [compute_distance((i + 1) % 3, i) for i in range(4)]
+        diagonals = [compute_distance(0, 2), compute_distance(1, 3)]
+        ratio = float(np.min(diagonals) / np.max(diagonals))
+
+        return [edges, diagonals, ratio]
+
+    @staticmethod
+    def _is_tile_distorted(self, corners: tuple) -> bool:
```
cdeil: Remove `self` argument here, it's never passed to `@staticmethod`s.
You then have to call `SimpleTilePainter._measure_tile_shape(corners)` instead of `self._measure_tile_shape(corners)`.
Overall my suggestion here would be to move these two functions just out of the `SimpleTilePainter` class, and make them standalone functions and put them above or below the `SimpleTilePainter` class.

adl1995:
@cdeil Alright, just made the changes. Please check.

I think that `i` would never reach to `4`? Since the loop is in `range(4)`, `i` would only be `0`, `1`, ` 2`, `3`, I think. However, I have changed it.

cdeil:
@adl1995 - FYI: I've done a small update to this code in 1b36234 
The small advantage of having the test this way is that it's independent and faster.

Another thing I noticed was that I had to put `bool` explicitly for `is_tile_distorted`. Previously it was returning `numpy.bool_` which has the very surprising behaviour:
```
>>> np.bool_(True) is True
False
```

Issue #80, opened 2017-07-19
Report to users what's going on


cdeil:
The current user experience with our package is pretty bad.

If I call `make_sky_image` to make a large sky image:
https://gist.github.com/cdeil/796875efe4a3620987b9f62eb278cfbc
it takes ~ a minute and I have no idea what's going on.

We should change our code to (optionally, at medium verbosity level by default) communicate to users what's happening:
- how many tiles / megabyte are being fetched
- when tile fetching starts
- when tile drawing starts
- how long fetching and drawing took
- how much memory was or is used

@adl1995 - I'm not sure how best to implement this yet. I guess we could start emitting log messages, or we could use a progress bar (http://docs.astropy.org/en/stable/api/astropy.utils.console.ProgressBarOrSpinner.html) ?

---

One thing I definitely think we should implement is a `HiPSDrawResult` class (or something like that), that is returned by `make_sky_image` instead of just the Numpy array.
It would have a `repr` that prints the info mentioned above (how long things took, how much memory was used, how much data was fetched from the web), and also has convenience methods to write and plot the resulting sky image:
```
result = make_sky_image(...)
result.report() # prints to console
result.show_image() # shows the image with matplotli
result.write_image('my_image.fits')` # Makes the HDU and writes it.
```

adl1995:
@cdeil I'm not entirely sure on how to go about implementing the `HipsDrawResult` class. One thing I though of is to introduce new methods in `SimpleTilePainter` for memory usage and data fetched and then call these from `HipsDrawResult`?

But, that doesn't seem like the best solution. I think the best way would be to contain all of the functionality within `HipsDrawResult` class, however, how would I get all the information from the like how much data is fetched, memory consumed, etc.?

cdeil:
Yes, eventually changes to the `SimpleTilePainter` will be needed, e.g. to compute these things.
But to start with, for the first pull request, please keep the changes to `SimpleTilePainter` to a minimum.

My suggestion would be that you introduce the new `HipsDrawResult` class, and change the
```
return painter.image
```
in `make_sky_image` to
```
return painter.result
```
and then implement a `result` property on the `SimpleTilePainter` that creates and returns a `HipsDrawResult` instance, based on already available info, starting with `image` and `geometry`.

[...]

PR #91, opened 2017-07-25
Introduce class HipsDrawResult in simple.py

adl1995:
As outlined in issue #80, I have added a class `HipsDrawResult`. This still needs more work, currently it only stores `image` and `geometry`.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -37,22 +37,34 @@ To make a sky image with the `hips` package, follow the following three steps:
 
 
 3. Call the `~hips.make_sky_image` function to fetch the HiPS data
-   and draw it, returning the sky image pixel data as a Numpy array::
+   and draw it, returning an object of `~hips.HipsDrawResult`::
 
     from hips import make_sky_image
 
-    data = make_sky_image(geometry, hips_survey, 'fits')
+    result = make_sky_image(geometry, hips_survey, 'fits')
 
 
 That's it. Go ahead and try it out for your favourite sky region and survey.
 
 Now you can then save the sky image to local disk e.g. FITS file format::
 
-    from astropy.io import fits
-    hdu = fits.PrimaryHDU(data=data, header=geometry.fits_header)
-    hdu.writeto('my_image.fits')
+    result.write_image('my_image.fits')
 
-or plot and analyse the sky image however you like.
+The ``result`` object also contains other useful information, such as::
+
+    result.image
+
+will return a NumPy array containing pixel data, you can also get the WCS information using::
+
+    result.geometry
+
+If you want, you could also print out information about the ``result``::
+
+    print(result)
+
+or plot and analyse the sky image using::
+
+    result.plot()
 
 If you execute the example above, you will get this sky image which was plotted using `astropy.visualization.wcsaxes`
 
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -1,6 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """HiPS tile drawing -- simple method."""
 import numpy as np
+from PIL import Image
+from astropy.io import fits
 from typing import List, Tuple
 from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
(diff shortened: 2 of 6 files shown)


cdeil:
This looks good to me. I think you can just continue with this PR for ~ a day, adding more features.

I would suggest adding a `__str__` method so that `print(result)` will print out infos about the result.
I've added `__repr__` to a few other other classes that do something similar, e.g. `HipsTileMeta`.
Here for the sky image, things that would be good to know are the width and height and number of channels, the dtype of the data, and some info about the WCS geometry.

Another thing I would suggest is to add a `write_image` method, so that users can do:
```
result.write_image('image.fits')
```
and then change the high-level docs example at https://hips.readthedocs.io/en/latest/getting_started.html to call that `write_image` convenience method instead of the three lines we currently have:
```
from astropy.io import fits
hdu = fits.PrimaryHDU(data=data, header=geometry.fits_header)
hdu.writeto('my_image.fits')
```

adl1995:
@cdeil I just made these updated. Please review.

hips/draw/tests/test_simple.py
```diff
@@ -46,13 +49,14 @@
...
-    assert_allclose(image[200, 994], pars['data_1'])
-    assert_allclose(image[200, 995], pars['data_2'])
-
+    result = make_sky_image(geometry=geometry, hips_survey=hips_survey, tile_format=pars['file_format'])
+    assert result.image.shape == pars['shape']
+    assert result.image.dtype == pars['dtype']
+    assert_allclose(np.sum(result.image), pars['data_sum'])
+    assert_allclose(result.image[200, 994], pars['data_1'])
+    assert_allclose(result.image[200, 995], pars['data_2'])
+    result.write_image('test.' + pars['file_format'])
```
cdeil: When writing files from tests, you should always do it using `tmpdir`.
Examples: https://github.com/hipspy/hips/search?utf8=%E2%9C%93&q=tmpdir&type=

hips/draw/simple.py
```diff
@@ -172,6 +179,53 @@ def plot_mpl_hips_tile_grid(self) -> None:
...
+        self.image = image
+        self.geometry = geometry
+        self.tile_format = tile_format
+
+    def __str__(self):
+        return (
+            'This class object contains two attributes: image and geometry'
+        )
+
+    def __repr__(self):
```
cdeil: Please follow the common pattern of mentioning the class name in the `repr`, like in this example:
https://github.com/hipspy/hips/blob/fce110c0e2bf31baba5d9050ccf08a9c67493dff/hips/tiles/tile.py#L99

hips/draw/simple.py
```diff
@@ -172,6 +179,53 @@ def plot_mpl_hips_tile_grid(self) -> None:
...
+    """
+
+    def __init__(self, image: np.ndarray, geometry: WCSGeometry, tile_format: str) -> None:
+        self.image = image
+        self.geometry = geometry
+        self.tile_format = tile_format
+
+    def __str__(self):
+        return (
+            'This class object contains two attributes: image and geometry'
```
cdeil: This isn't useful. The idea of having a separate `__str__` is that you give more information.
In this case, it could be something like this:
```
def __str__(self):
    ss = 'HiPS draw result:\n'
    ss += 'Sky image: .... infos about sky image pixel data (shape, dtype) ...\n'
    ss += f'WCS geometry: {self.geometry}\n'
    .... later add more info about number of tiles used and fetched, time to exectute, memory used
   ... i.e. print out a nice report
    return ss
```

docs/getting_started.rst
```diff
@@ -37,20 +37,18 @@ To make a sky image with the `hips` package, follow the following three steps:
...
 
 
 That's it. Go ahead and try it out for your favourite sky region and survey.
 
 Now you can then save the sky image to local disk e.g. FITS file format::
 
-    from astropy.io import fits
-    hdu = fits.PrimaryHDU(data=data, header=geometry.fits_header)
-    hdu.writeto('my_image.fits')
+    result.write_image('my_image.fits')
```
cdeil: This info in the high-level section is now a little short, you should give some more infos what people can do with the `result`, namely that they can get at the pixel data as a numpy array, the WCS for the image and print out info about the result.

adl1995:
@cdeil I just pushed my latest code. Now there is a method in `HipsDrawResult` named `plot` which plots the all-sky image with HEALPix grid. I also made some changes to the high-level docs and added the `__str__` method in `WCSGeometry`.

cdeil:
Thanks! (I'll do some small follow-up edits in master later today.)

PR #94, opened 2017-07-27
Update painter, survey from name and getting started

cdeil:
@adl1995 - This PR is partly the result of the pair coding we did yesterday.

It contains:
- Update of getting started guide
- Introduce HiPS survey lookup by name (just outlined, no tests or docs yet)
- Rename `SimpleTilePainter` to `HipsPainter`, because we're adding the distortion correction to it also. (We'll have to discuss when we're all back from vacation whether we want to drop the "Hips" prefix from all of our class names in this package, and if the overall names are good and clear.)
- Misc small code and docstring (hopefully) improvements

I will merge this if travis-ci tests pass.

### DIFF
--- a/docs/drawing_algo.rst
+++ b/docs/drawing_algo.rst
@@ -19,7 +19,7 @@ Naive algorithm
 ===============
 
 This is a naive (one could also say: simple and fast) algorithm for drawing HiPS
-tiles using affine transformations, implemented in the `~hips.SimpleTilePainter`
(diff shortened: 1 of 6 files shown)


cdeil:
@adl1995 - I'm merging this now so that you can continue, even though I'm not sure if this test is hanging on travis-ci or not: https://travis-ci.org/hipspy/hips/jobs/258078306#L1227
(works for me locally)

Issue #82, opened 2017-07-19
Introduce a simple way to specify WCSGeometry


cdeil:
The current method we have to specify `WCSGeometry` is pretty complicated:
* https://hips.readthedocs.io/en/latest/getting_started.html#make-a-sky-image
* https://hips.readthedocs.io/en/latest/api/hips.WCSGeometry.html
```
from astropy.coordinates import SkyCoord
from hips.utils import WCSGeometry
geometry = WCSGeometry.create(
     skydir=SkyCoord(0, 0, unit='deg', frame='galactic'),
     width=2000, height=1000, fov="3 deg",
     coordsys='galactic', projection='AIT',
)
```
---

In contrast, Aladin Lite and ipyaladin have this nice simple way:
https://github.com/cds-astro/ipyaladin/blob/master/examples/2_Base_Commands.ipynb

---

The main difference is that they have `target` which can be a string like "Crab nebula" that's looked up via SESAME, or it can be a string representation of the sky coordinate like "0. 0.", whereas we currently have to import SkyCoord and create it, which means two extra lines.

There is 
http://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.from_name to look up from SESAME, I don't know if they have a Skycoord string parsing functionality???

So I think we should also access `skydir` strings (maybe rename to `target`) and the simple interface.

In addition, we might want to allow the `geometry` parameter in `make_sky_image` to be a dict, and if it is, call `WCSGeometry.create_from_dict()` for the user as a convenience. I.e. this would work:
```
geometry = dict(target='crab', fov='10 deg', frame='galactic')
make_sky_image(geometry, hips_survey)
```

cdeil:
I see that this issue was partly implemented by @adl1995  in #100. 

The main missing pieces are to
- use this simpler from dict method in the high-level docs getting started example (and `plot_*.py` files in `docs`)
- make it possible / easier for someone looking at the `make_sky_image` docstring to figure out what options they can pass, i.e. link from there to the other docstring of the method that is called.

cdeil:
@tboch - I think in terms of convenience API to define common WCS in a simple way, we might want to just re-use the one from https://github.com/cds-astro/ipyaladin  and get rid of our `WCSGeometry`, or at least add a classmethod to suport what ipyaladin does?

Would be nice if those two Python HiPS packages had a similar API, no?

[...]

PR #95, opened 2017-07-27
Add option to create HipsSurveyProperties from survey name

adl1995:
As outlined in #82 and during the telecon yesterday, I have added the option to create `HipsSurveyProperties` using a survey name.

@cdeil Currently, this is not implemented:

> 
> In addition, we might want to allow the geometry parameter in make_sky_image to be a dict, and if it is, call WCSGeometry.create_from_dict() for the user as a convenience. I.e. this would work:
> 
> ```
> geometry = dict(target='crab', fov='10 deg', frame='galactic')
> make_sky_image(geometry, hips_survey)
> 
> ```

Can this be part of a separate PR?

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -26,21 +26,16 @@ To draw a sky image from HiPS image tiles with the `hips` package, follow the fo
          coordsys='galactic', projection='AIT',
     )
 
-
-2. Specify the HiPS survey you want by creating a `~hips.HipsSurveyProperties` object.
+2. Specify the HiPS survey you want. You just need to provide a valid HiPS survey ID.
 
    A good address that lists available HiPS data is http://aladin.u-strasbg.fr/hips/list ::
 
-    from hips import HipsSurveyProperties
-    url = 'http://alasky.unistra.fr/DSS/DSS2Merged/properties'
-    hips_survey = HipsSurveyProperties.fetch(url)
-
+    hips_survey = 'CDS/P/DSS2/red'
 
 3. Call the `~hips.make_sky_image` function to fetch the HiPS data
    and draw it, returning an object of `~hips.HipsDrawResult`::
 
     from hips import make_sky_image
-
     result = make_sky_image(geometry, hips_survey, 'fits')
 
 Of course, you could change the parameters to chose any sky image geometry and
--- a/hips/draw/simple.py
+++ b/hips/draw/simple.py
@@ -4,7 +4,7 @@ import time
 import numpy as np
 from PIL import Image
 from astropy.io import fits
-from typing import List, Tuple
+from typing import List, Tuple, Union
 from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
 from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta
@@ -61,7 +61,7 @@ class HipsPainter:
     (1000, 2000)
     """
 
-    def __init__(self, geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile_format: str) -> None:
+    def __init__(self, geometry: WCSGeometry, hips_survey: Union[str, HipsSurveyProperties], tile_format: str) -> None:
         self.geometry = geometry
         self.hips_survey = HipsSurveyProperties.make(hips_survey)
         self.tile_format = tile_format
@@ -382,7 +382,7 @@ def plot_mpl_single_tile(geometry: WCSGeometry, tile: HipsTile, image: np.ndarra
     ax.imshow(image, origin='lower')
 
 
-def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile_format: str) -> 'HipsDrawResult':
+def make_sky_image(geometry: WCSGeometry, hips_survey: Union[str, 'HipsSurveyProperties'], tile_format: str) -> 'HipsDrawResult':
     """Make sky image: fetch tiles and draw.
 
     The example for this can be found on the :ref:`gs` page.
@@ -399,8 +399,8 @@ def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties, til
 
     Returns
     -------
-    image : `~numpy.ndarray`
-        Output image pixels
+    result : `~hips.HipsDrawResult`
+        Result object
     """
     painter = HipsPainter(geometry, hips_survey, tile_format)
     painter.run()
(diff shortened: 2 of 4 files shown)


docs/getting_started.rst
```diff
@@ -26,21 +26,17 @@ To draw a sky image from HiPS image tiles with the `hips` package, follow the fo
...
-
 2. Specify the HiPS survey you want by creating a `~hips.HipsSurveyProperties` object.
 
    A good address that lists available HiPS data is http://aladin.u-strasbg.fr/hips/list ::
 
     from hips import HipsSurveyProperties
-    url = 'http://alasky.unistra.fr/DSS/DSS2Merged/properties'
-    hips_survey = HipsSurveyProperties.fetch(url)
-
+    hips_survey = HipsSurveyProperties.make('CDS/P/DSS2/red')
```
cdeil: If you remove the import and just write this here
```
hips_survey = 'CDS/P/DSS2/red'
```
it would give the same result, no?

Please make it work that way and change the high-level docs to use that simple form.

hips/draw/simple.py
```diff
@@ -61,7 +61,7 @@ class HipsPainter:
     (1000, 2000)
     """
 
-    def __init__(self, geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile_format: str) -> None:
+    def __init__(self, geometry: WCSGeometry, hips_survey: str or HipsSurveyProperties, tile_format: str) -> None:
```
cdeil: I think the type annotation has to be `Union` in case of multiple options, no?
https://docs.python.org/3/library/typing.html#typing.Union

adl1995:
@cdeil I've made the changes. I also marked the tests with `remote_data`. Hopefully, the build will pass now.

docs/getting_started.rst
```diff
@@ -26,21 +26,16 @@ To draw a sky image from HiPS image tiles with the `hips` package, follow the fo
          coordsys='galactic', projection='AIT',
     )
 
-
-2. Specify the HiPS survey you want by creating a `~hips.HipsSurveyProperties` object.
+2. Specify the HiPS survey you want.
```
cdeil: You should mention that users have to give the ID of the Hips survey.

hips/draw/simple.py
```diff
@@ -35,7 +35,7 @@ class HipsPainter:
     ----------
     geometry : `~hips.utils.WCSGeometry`
         An object of WCSGeometry
-    hips_survey : str or `~hips.HipsSurveyProperties`
+    hips_survey : `str` or `~hips.HipsSurveyProperties`
```
cdeil: I'm using the docs pattern where I don't put backticks for PYthon built-in types like `str`, `list`, `int`, ...
So please remove them here for consistency with the rest of our docstrings.

hips/draw/simple.py
```diff
@@ -61,7 +61,7 @@ class HipsPainter:
     (1000, 2000)
     """
 
-    def __init__(self, geometry: WCSGeometry, hips_survey: HipsSurveyProperties, tile_format: str) -> None:
+    def __init__(self, geometry: WCSGeometry, hips_survey: Union[str, 'HipsSurveyProperties'], tile_format: str) -> None:
```
cdeil: The common thing to do is to put the type, and only to put quotes if there's a forward reference issue.
So please remove the quotes from `HipsSurveyProperties` here in the type annotation, and review the diff of this PR on Github to fix the other cases (I see one below).

hips/draw/simple.py
```diff
@@ -391,16 +391,16 @@ def make_sky_image(geometry: WCSGeometry, hips_survey: HipsSurveyProperties, til
...
     tile_format : {'fits', 'jpg', 'png'}
         Format of HiPS tile to use
         (some surveys are available in several formats, so this extra argument is needed)
 
     Returns
     -------
-    image : `~numpy.ndarray`
-        Output image pixels
+    result : `~hips.HipsDrawResult`
+        HipsDrawResult object
```
cdeil: Don't duplicate the type name in the description. It's not useful. Here I'd suggest this desription:
```
    Result object
```

hips/tiles/tests/test_surveys.py
```diff
@@ -85,8 +96,6 @@ def test_fetch(self):
         surveys = HipsSurveyPropertiesList.fetch()
         assert len(surveys.data) > 3
 
-        # TODO: look up survey by name here
-        # Otherwise this will break when the list changes
-        survey = surveys.data[0]
+        survey = HipsSurveyProperties.from_name('CDS/P/2MASS/H')
```
cdeil: This change to the test is no good.

I would suggest to move the code that does the name lookup from the `HipsSurveyProperties.from_name` method to a `HipsSurveyProperties.from_name` method and then call this here from this test (instead of calling
```
HipsSurveyProperties.from_name('CDS/P/2MASS/H')
```
which will fetch the list again.

The implementation of `HipsSurveyProperties.from_name` would then just be
```python
@classmethod
def from_name(name);
    surveys = HipsSurveyProperties.fetch()
    return surveys.from_name(name)
```
and maybe we'd remove it eventually.

You should also add a test with an invalid key like `name='kronka lonka'` and check that a KeyError is raised.
I explained how to do that with `pytest.raises` here previously: https://github.com/hipspy/hips/pull/55#discussion_r125573377

hips/tiles/tests/test_surveys.py
```diff
@@ -11,33 +11,44 @@ class TestHipsSurveyProperties:
...
+    @remote_data
+    def test_from_name(self):
+        survey = HipsSurveyProperties.from_name('CDS/P/2MASS/color')
+        assert survey.title == '2MASS color J (1.23 microns), H (1.66 microns), K (2.16 microns)'
+
+    @remote_data
+    def test_make(self):
+        survey = HipsSurveyProperties.make('CDS/P/EGRET/Dif/300-500')
+        assert survey.title == 'EGRET Dif 300-500'
+        self.survey = HipsSurveyProperties.make(self.survey)
```
cdeil: This line doesn't make sense. You're setting `self.survey` on the last line of the test.
Instead you should be asserting something.
I think this might work:
```
assert self.survey is HipsSurveyProperties.make(self.survey)
```
because in that case we just return the same object and "is" checks for object equality, no?

hips/tiles/tests/test_surveys.py
adl1995: @cdeil I'm not sure what you mean here. Do you mean to say create a simple class method `from_name` in addition to the `classmethod`?

cdeil: I've updated my comment above: the class names I gave were wrong. I meant adding a normal method `HipsSurveyPropertiesList.from_name` (not a classmethod), because really looking for a name is an operation on the list, i.e. the code that does the looping through surveys and looking for a name ID should be a method on `HipsSurveyPropertiesList`.

Is it clear now?

cdeil:
Thanks!

Issue #98, opened 2017-07-31
Incorrect JPG, PNG precise tile drawing


adl1995:
I'm currently working on introducing precise drawing, the tile splitting works correctly, but, when I try and draw the all sky image, I end up with something like this:

![precise-drawing](https://user-images.githubusercontent.com/13198325/28770623-4c0a95e8-75f9-11e7-8e40-864e40cb2e21.png)

But, it works fine using the simple algorithm:

![simple-drawing](https://user-images.githubusercontent.com/13198325/28770638-58cf4b98-75f9-11e7-89e8-f587c3ca87c9.png)

Note: this issue is only for JPG and PNG tiles, the FITS tiles are being drawn correctly using the precise algorithm.

I tried displaying the JPG tile children individually, and I get this result:

![four-child-tiles](https://user-images.githubusercontent.com/13198325/28770687-812d7ace-75f9-11e7-9e90-8449cab9db3e.png)
 
The original tile is this:

![single-tile](https://user-images.githubusercontent.com/13198325/28770695-863edfd0-75f9-11e7-8d04-4d43f0d4d396.png)

So, there is no issue with the `children` property as far as I can tell, my guess is that this has something to do with how the corners are being computed.

Since @cdeil is on vacation, @tboch can you please take a look at this?

tboch:
yes, I'll have a look at this ASAP.

tboch:
@adl1995 : as JPEG/PNG tiles are vertically flipped with respect to FITS tiles, I think that the order of items in the `children` property has to be changed: for JPEG/PNG tiles, items at indices 0 and 1 must be swapped, same thing for items at indices 2 and 3.
Can you try this and report if it works fine?

adl1995:
@tboch Did you mean changing the order of `children` tiles? You can see the updated code here: https://github.com/adl1995/hips/blob/precise/hips/tiles/tile.py#L252

But, this did not have any effect. Here is the updated notebook: https://github.com/adl1995/HIPS-to-Py/blob/master/examples/Precise%20drawing.ipynb

Also, the order of children tiles is now incorrect, as it can be seen in the last cell of the notebook.

tboch:
yes, that's what I meant.
Let me have a closer look at this.

tboch:
@adl1995 OK, I think I understand better now what happens now. Actually, when children tiles are created, the order of items in the `data` array must be changed as follow:
```python
        data = [
            self.data[w: w * 2, 0: w],
[...]

PR #99, opened 2017-08-01
Introduce precise drawing algorithm

adl1995:
As discussed in #98 and #84, I have introduced the precise drawing using a boolean named `precise`, which defaults to `False` for now. The result can be seen on this notebook: [https://github.com/adl1995/HIPS-to-Py/blob/master/examples/Precise drawing.ipynb](https://github.com/adl1995/HIPS-to-Py/blob/master/examples/Precise%20drawing.ipynb)

I have copied the resulting images below as well. Precise drawing:

precise

Simple drawing:

simple

Although the results are not apparent here, they can clearly be seen when viewed through an image viewer. Also currently there is no recursion involved. @tboch Please take a look.


cdeil:
🎉

PR #100, opened 2017-08-02
Add classmethod create_from_dict in WCSGeometry class

adl1995:
As @cdeil mentioned in #82, I have added a classmethod `create_from_dict` in `WCSGeometry` class. Now the user can pass either a `dict` or a `WCSGeometry` object itself. There is also a convenience method `make` in `WCSGeometry` which creates the object from `dict` parameters, if required.

### DIFF
--- a/hips/utils/tests/test_wcs.py
+++ b/hips/utils/tests/test_wcs.py
@@ -34,3 +34,11 @@ class TestWCSGeometry:
         assert_allclose(c.dec.deg, 0.00075, atol=1e-2)
         assert_allclose(self.geometry.wcs.wcs.crpix, [1000., 500.])
         assert_allclose(self.geometry.wcs.wcs.cdelt, [-0.0015, 0.0015])
+
+    def test_create_from_dict(self):
+        params = dict(target='crab', width=2000, height=1000, fov='3 deg',
+                      coordsys='galactic', projection='AIT')
+        geometry = WCSGeometry.create_from_dict(params)
+
+        c = self.geometry.center_skycoord
+        assert c.frame.name == 'galactic'
--- a/hips/utils/tests/test_wcs.py
+++ b/hips/utils/tests/test_wcs.py
@@ -40,5 +40,9 @@ class TestWCSGeometry:
                       coordsys='galactic', projection='AIT')
         geometry = WCSGeometry.create_from_dict(params)
 
-        c = self.geometry.center_skycoord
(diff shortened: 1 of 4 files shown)


hips/utils/tests/test_wcs.py
```diff
@@ -34,3 +34,11 @@ def test_celestial_frame(self):
...
         assert_allclose(self.geometry.wcs.wcs.crpix, [1000., 500.])
         assert_allclose(self.geometry.wcs.wcs.cdelt, [-0.0015, 0.0015])
+
+    def test_create_from_dict(self):
+        params = dict(target='crab', width=2000, height=1000, fov='3 deg',
+                      coordsys='galactic', projection='AIT')
+        geometry = WCSGeometry.create_from_dict(params)
+
+        c = self.geometry.center_skycoord
+        assert c.frame.name == 'galactic'
```
tboch: I think we could test a couple of other properties here, don't you think?

adl1995:
@tboch I've made the changes, however, Travis CI build is failing. I don't know what's causing this, the line reads: https://travis-ci.org/hipspy/hips/jobs/260162492#L1260

> INTERNALERROR> AttributeError: 'NoneType' object has no attribute 'lineno'

The tests do pass locally.

tboch:
Thanks for letting me know about the failing test, I was about to ask.

Merging now

cdeil:
Concerning the error you mentioned above, it's not related to our `hips` package.
I'll try to follow up and get it fixed:
https://github.com/astropy/astropy/issues/5277#issuecomment-319956530

Locally I see this fail: https://gist.github.com/cdeil/e1383b1270d32c12a87d7954958f9b82
@adl1995 - Can you please fix in a follow-up PR, and also remove the commented out `print` statement?

I'd also suggest to improve the docstring for this new method. There must be a way for a user to figure out which options they can pass, no? For now, adding a sentence saying that extra options are forwarded to `create` would be useful, i.e. users can see the list of options there.

One more thing: in http://hips.readthedocs.io/en/latest/api/hips.WCSGeometry.html#hips.WCSGeometry.create you can see that the link to
```
fov: str or `~astropy.coordinates.Angle`
```
is broken. The cause is a missing space before the colon, it should be
```
fov : ...
```
@adl1995 - Can you please fix that as well in the the same follow-up PR, and generally fix and improve such little things whenever you notice them?

Issue #92, opened 2017-07-25
Tile splitting diagonal length criterion


cdeil:
@tboch - At https://github.com/hipspy/hips/blame/71e593ab7e60767be70d9b2b13398016c35db09a/docs/drawing_algo.rst#L73 you wrote

> To overcome this problem, Aladin Desktop and Aladin Lite use the following recursive strategy: for tiles either too large (one edge is >300 pixels or diagonal is > 150 pixels when projected) or too distorted (ratio of smaller diagonal on larger diagonal is smaller than 0.7)

and by now @adl1995 has implemented that criterion in a helper function in `draw.py`.

I think maybe that is a typo and really the diagonal length limit should be 350 or 450 pix instead of 150 as written?
My reasoning is that one of the two diagonals is **always** longer than any of the edges, so as written the "one edge is > 300 pixels" criterion is superfluous, because the "diagonal is > 150 pix" criterion would already always be true in those cases, no?

tboch:
Actually, after re-checking, the criterion to use to determine if a file is too large is as follow:

- either one edge is > 300 pixels when projected
- or the tile is considered as distorted (ratio of smaller diagonal on larger diagonal < 0.7) AND one of the diagonal is > 150 pixels when projected

cdeil:
@adl1995 - Please update the implementation (and possibly tests) to match the description.

adl1995:
Fixed in #101.

PR #101, opened 2017-08-07
Fix tile splitting criterion

adl1995:
The tile splitting criterion is updated following this comment: https://github.com/hipspy/hips/issues/92#issuecomment-319923594


cdeil:
Looks good to me. Thanks!

PR #102, opened 2017-08-08
Reformat docstring for create_from_dict class method

adl1995:
As @cdeil mentioned in #100: https://github.com/hipspy/hips/pull/100#issuecomment-319957989, I have made small improvements to the docstring, apart from some other changes.

### DIFF
--- a/hips/tiles/tests/test_allsky.py
+++ b/hips/tiles/tests/test_allsky.py
@@ -11,7 +11,7 @@ TEST_CASES = [
     dict(
         label='fits',
         meta=dict(order=3, ipix=-1, file_format='fits'),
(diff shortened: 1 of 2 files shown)


adl1995:
@cdeil - I'm not sure why the test fails at this line:

`skycoord = SkyCoord.from_name(params.pop('target'), frame=params['coordsys'])`

In this case, `target = 'crab'`, but the test does pass locally.

cdeil:
Looks good to me. Merging now.

Concerning the `target = 'crab'` test fail: let's just merge everything in and then see if there's still a test fail locally or on travis-ci (and if so, fix it in a follow-up PR).

PR #104, opened 2017-08-09
Add reporting functionality for HipsPainter

adl1995:
I have added a minimal functionality for the `HipsPainter` class. For now, the following things are reported:

- Fetch time
- Draw time
- Total memory consumed
- Total tiles fetched


cdeil:
Thanks!

PR #106, opened 2017-08-15
Introduce asynchronous fetching of HiPS tiles

adl1995:
As the title states, HiPS tiles are now fetched asynchronously. I noted 4 seconds for tile fetching which previously took around 25 seconds at my end. For now, this is introduced in the `HipsPainter` class, but let me know if this needs to be refactored somewhere else.

### DIFF
--- a/hips/draw/paint.py
+++ b/hips/draw/paint.py
@@ -37,7 +37,8 @@ class HipsPainter:
     progress_bar : bool
         Show a progress bar for tile fetching and drawing
     fetch_opts : dict
-        Keyword arguments for fetching HiPS tiles
+        Keyword arguments for fetching HiPS tiles. To see the
+        list of passable arguments, refer to `~fetch_tiles`
 
     Examples
     --------
@@ -127,7 +128,7 @@ class HipsPainter:
 
         if self._tiles is None:
             self._tiles = fetch_tiles(tile_metas=tile_metas, hips_survey=self.hips_survey,
-                                      progress_bar=self.progress_bar, **self.fetch_opts)
+                                      progress_bar=self.progress_bar, **(self.fetch_opts or {}))
 
         return self._tiles
 
--- a/hips/draw/paint.py
+++ b/hips/draw/paint.py
@@ -6,7 +6,7 @@ import numpy as np
 from typing import List, Tuple, Union, Dict, Any, Generator
 from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
-from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta
+from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta, HipsTileFetcher
 from ..tiles.tile import compute_image_shape
 from ..utils import WCSGeometry, healpix_pixels_in_sky_image, hips_order_for_pixel_resolution
 
@@ -153,7 +153,8 @@ class HipsPainter:
     def tiles(self) -> List[HipsTile]:
         """List of `~hips.HipsTile` (cached on multiple access)."""
         if self._tiles is None:
-            self._tiles = asyncio.get_event_loop().run_until_complete(self._fetch_tiles())
+            # self._tiles = asyncio.get_event_loop().run_until_complete(tile_fetcher.tiles)
+            self._tiles = tile_fetcher.tiles
 
         return self._tiles
 
--- a/hips/draw/paint.py
+++ b/hips/draw/paint.py
@@ -4,7 +4,7 @@ import numpy as np
 from typing import List, Tuple, Union, Dict, Any
 from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
-from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta, HipsTileFetcher
+from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta, fetch_tiles
 from ..tiles.tile import compute_image_shape
 from ..utils import WCSGeometry, healpix_pixels_in_sky_image, hips_order_for_pixel_resolution
 
@@ -36,8 +36,8 @@ class HipsPainter:
         Use the precise drawing algorithm
     progress_bar : bool
         Show a progress bar for tile fetching and drawing
-    fetch_package : {'urllib', 'aiohttp'}
-        Package to use for fetching HiPS tiles
+    fetch_opts : dict
+        Keyword arguments for fetching HiPS tiles
 
     Examples
     --------
@@ -61,13 +61,13 @@ class HipsPainter:
     """
 
     def __init__(self, geometry: Union[dict, WCSGeometry], hips_survey: Union[str, HipsSurveyProperties],
-                 tile_format: str, precise: bool = False, progress_bar: bool = True, fetch_package: str = 'urllib') -> None:
+                 tile_format: str, precise: bool = False, progress_bar: bool = True, fetch_opts : dict = None) -> None:
         self.geometry = WCSGeometry.make(geometry)
         self.hips_survey = HipsSurveyProperties.make(hips_survey)
         self.tile_format = tile_format
         self.precise = precise
         self.progress_bar = progress_bar
-        self.fetch_package = fetch_package
+        self.fetch_opts = fetch_opts
         self._tiles = None
         self.float_image = None
         self._stats: Dict[str, Any] = {}
@@ -125,11 +125,9 @@ class HipsPainter:
             )
             tile_metas.append(tile_meta)
 
-        tile_fetcher = HipsTileFetcher(tile_metas=tile_metas, hips_survey=self.hips_survey,
-                                       progress_bar=self.progress_bar, fetch_package=self.fetch_package)
-
         if self._tiles is None:
-            self._tiles = tile_fetcher.tiles
+            self._tiles = fetch_tiles(tile_metas=tile_metas, hips_survey=self.hips_survey,
+                                      progress_bar=self.progress_bar, **self.fetch_opts)
 
         return self._tiles
 
--- a/hips/draw/paint.py
+++ b/hips/draw/paint.py
@@ -1,7 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import time
+import aiohttp
+import asyncio
 import numpy as np
-from typing import List, Tuple, Union, Dict, Any, Iterator
+from typing import List, Tuple, Union, Dict, Any, Generator
 from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
 from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta
@@ -109,7 +111,12 @@ class HipsPainter:
         pt.estimate(src, dst)
         return pt
 
-    def _fetch_tiles(self) -> Iterator[HipsTile]:
+    async def fetch_tile_threaded(self, url: str, session: aiohttp.client.ClientSession) -> Generator:
+        """Fetch a HiPS tile asynchronously."""
+        async with session.get(url) as response:
+            return await response.read()
+
+    async def _fetch_tiles(self) -> List[HipsTile]:
         """Generator function to fetch HiPS tiles from a remote URL."""
         if self.progress_bar:
             from tqdm import tqdm
@@ -118,21 +125,35 @@ class HipsPainter:
             tile_indices = self.tile_indices
 
         for healpix_pixel_index in tile_indices:
+        tile_urls, tile_metas = [], []
+        for healpix_pixel_index in self.tile_indices:
             tile_meta = HipsTileMeta(
                 order=self.draw_hips_order,
                 ipix=healpix_pixel_index,
                 frame=self.hips_survey.astropy_frame,
                 file_format=self.tile_format,
             )
-            url = self.hips_survey.tile_url(tile_meta)
-            tile = HipsTile.fetch(tile_meta, url)
-            yield tile
+            tile_urls.append(self.hips_survey.tile_url(tile_meta))
+            tile_metas.append(tile_meta)
+
+        tasks = []
+        async with aiohttp.ClientSession() as session:
+            for idx, url in enumerate(tile_urls):
+                task = asyncio.ensure_future(self.fetch_tile_threaded(url.format(idx), session))
+                tasks.append(task)
+
+            raw_responses = await asyncio.gather(*tasks)
+
+        tiles = []
+        for idx, raw_data in enumerate(raw_responses):
+            tiles.append(HipsTile(tile_metas[idx], raw_data))
+        return tiles
 
     @property
     def tiles(self) -> List[HipsTile]:
         """List of `~hips.HipsTile` (cached on multiple access)."""
         if self._tiles is None:
-            self._tiles = list(self._fetch_tiles())
+            self._tiles = asyncio.get_event_loop().run_until_complete(self._fetch_tiles())
 
         return self._tiles
 
@@ -161,7 +182,6 @@ class HipsPainter:
         self._stats['consumed_memory'] = 0
         for tile in self.draw_tiles:
             self._stats['consumed_memory'] += len(tile.raw_data)
-
 
 
     def make_tile_list(self):
--- a/hips/draw/paint.py
+++ b/hips/draw/paint.py
@@ -1,9 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import time
-import aiohttp
-import asyncio
 import numpy as np
-from typing import List, Tuple, Union, Dict, Any, Generator
+from typing import List, Tuple, Union, Dict, Any
 from astropy.wcs.utils import proj_plane_pixel_scales
 from skimage.transform import ProjectiveTransform, warp
 from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta, HipsTileFetcher
@@ -38,6 +36,8 @@ class HipsPainter:
         Use the precise drawing algorithm
     progress_bar : bool
         Show a progress bar for tile fetching and drawing
+    fetch_package : {'urllib', 'aiohttp'}
+        Package to use for fetching HiPS tiles
 
     Examples
     --------
@@ -61,12 +61,13 @@ class HipsPainter:
     """
 
     def __init__(self, geometry: Union[dict, WCSGeometry], hips_survey: Union[str, HipsSurveyProperties],
-                 tile_format: str, precise: bool = False, progress_bar: bool = True) -> None:
+                 tile_format: str, precise: bool = False, progress_bar: bool = True, fetch_package: str = 'urllib') -> None:
         self.geometry = WCSGeometry.make(geometry)
         self.hips_survey = HipsSurveyProperties.make(hips_survey)
         self.tile_format = tile_format
         self.precise = precise
         self.progress_bar = progress_bar
+        self.fetch_package = fetch_package
         self._tiles = None
         self.float_image = None
         self._stats: Dict[str, Any] = {}
@@ -111,21 +112,10 @@ class HipsPainter:
         pt.estimate(src, dst)
         return pt
 
-    async def fetch_tile_threaded(self, url: str, session: aiohttp.client.ClientSession) -> Generator:
-        """Fetch a HiPS tile asynchronously."""
-        async with session.get(url) as response:
-            return await response.read()
-
-    async def _fetch_tiles(self) -> List[HipsTile]:
-        """Generator function to fetch HiPS tiles from a remote URL."""
-        if self.progress_bar:
-            from tqdm import tqdm
-            tile_indices = tqdm(self.tile_indices, desc='Fetching tiles')
-        else:
-            tile_indices = self.tile_indices
-
-        for healpix_pixel_index in tile_indices:
-        tile_urls, tile_metas = [], []
+    @property
+    def tiles(self) -> List[HipsTile]:
+        """List of `~hips.HipsTile` (cached on multiple access)."""
+        tile_metas = []
         for healpix_pixel_index in self.tile_indices:
             tile_meta = HipsTileMeta(
                 order=self.draw_hips_order,
@@ -133,27 +123,12 @@ class HipsPainter:
                 frame=self.hips_survey.astropy_frame,
                 file_format=self.tile_format,
             )
-            tile_urls.append(self.hips_survey.tile_url(tile_meta))
             tile_metas.append(tile_meta)
 
-        tasks = []
-        async with aiohttp.ClientSession() as session:
-            for idx, url in enumerate(tile_urls):
-                task = asyncio.ensure_future(self.fetch_tile_threaded(url.format(idx), session))
-                tasks.append(task)
+        tile_fetcher = HipsTileFetcher(tile_metas=tile_metas, hips_survey=self.hips_survey,
+                                       progress_bar=self.progress_bar, fetch_package=self.fetch_package)
 
-            raw_responses = await asyncio.gather(*tasks)
-
-        tiles = []
-        for idx, raw_data in enumerate(raw_responses):
-            tiles.append(HipsTile(tile_metas[idx], raw_data))
-        return tiles
-
-    @property
-    def tiles(self) -> List[HipsTile]:
-        """List of `~hips.HipsTile` (cached on multiple access)."""
         if self._tiles is None:
-            # self._tiles = asyncio.get_event_loop().run_until_complete(tile_fetcher.tiles)
             self._tiles = tile_fetcher.tiles
 
         return self._tiles
--- a/hips/tiles/fetch.py
+++ b/hips/tiles/fetch.py
@@ -1,17 +1,22 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-import numpy as np
+import asyncio
 import urllib.request
 import concurrent.futures
 from typing import Generator, List
 from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta
 
 __all__ = [
-    'HipsTileFetcher',
+    'fetch_tiles',
 ]
 
-
-class HipsTileFetcher:
+def fetch_tiles(tile_metas: List[HipsTileMeta], hips_survey : HipsSurveyProperties,
+                progress_bar: bool = False, n_parallel: int = 10,  timeout: int = 10, fetch_package : str = 'urllib') -> List[HipsTile]:
     """Fetch a list of HiPS tiles.
+
+    This function fetches a list of HiPS tiles based
+    on their URLs, which are generated using `hips_survey`
+    and `tile_metas`. The tiles are then fetched asynchronously
+    using urllib or aiohttp.
 
     Parameters
     ----------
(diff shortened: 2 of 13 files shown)


cdeil:
@adl1995 - Thanks!

Please put this on the list of things to discuss in the telcon this afternoon.

For sure, this code shouldn't be located in `hips/draw/paint.py`.

As discussed since the start of the project (and started by you in #33), the tile fetching / caching should be in it's own file. We've discussed it many times, but it hasn't happened so far, we should really make it a priority for the last week.

One reason that it's difficult is that (I think) there's a fundamental decision to be made whether we are happy with two phases that happen after each other (first fetch all tiles, then do all drawing), or whether we want to support async tile fetching and immediately and async tile drawing as the tiles come it. That second option is faster, and also much more memory-efficient (can delete tiles that are drawn from memory immediately), but also leads to tighter coupling between the fetching / caching and drawing code. My suggestion for now (and I think has always been) would be that we implement the simple solution and move the tile fetching out to a separate Python file, and have it as much as possible decoupled from the drawing, i.e. the drawer just gives a list of tiles needed to the fetcher and is handed back that list of tiles, i.e. there's two lines of code in the painter calling the fetcher.

---

The second question / suggestion I have is whether introducing `aiohttp` as a dependency is really needed, or whether an async or at least parallel via threads fetch wouldn't work for our simple application and we would avoid the extra dependency. There's e.g. this:
https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example

@adl1995 - I know you have looked at this before GSoC started. Please remind us of the results this afternoon and then we discuss.

Also: if we go the simple route of separating out the fetching, then we could simply have two different fetchers, one using `aiohttp` as an optional dependency and the other just the Python standard lib to fetch, and the user could select via a simple option which one to use (probably defaulting to the Python standard lib, and only offering `aiohttp` as opt-in for power-users that really need the extra little bit of performance.

adl1995:
@cdeil I have added asynchronous tile fetching using both `urllib` and `aiohttp`, however, the test fails at:
https://travis-ci.org/hipspy/hips/jobs/265656774

I think this is due to the reason that tiles are not fetched sequentially and when they are matched with `HipsTileMeta`, they are misaligned. [https://github.com/hipspy/hips/pull/106/files#diff-27bdb9fa08176e9f36aa86205b78af6eR71](https://github.com/hipspy/hips/pull/106/files#diff-27bdb9fa08176e9f36aa86205b78af6eR71)

What do you think is the best solution to this? Should I somehow label each tile, so I can sort them once they are fetched?

cdeil:
This PR has changes to `astropy_helpers` which shouldn't be here. Please fix.

hips/draw/paint.py
```diff
@@ -36,6 +36,8 @@ class HipsPainter:
         Use the precise drawing algorithm
     progress_bar : bool
         Show a progress bar for tile fetching and drawing
+    use_aiohttp : bool
```
cdeil: I think a multiple-choice string option would be better, for two reasons:
1. it's extensible, in case we want to add a third implementation
2. it's clearer; with `use_aiohttp=False` and your parameter description, the user has no idea what they get.

hips/draw/paint.py
```diff
@@ -109,30 +112,17 @@ def projection(self, tile: HipsTile) -> ProjectiveTransform:
...
     @property
     def tiles(self) -> List[HipsTile]:
         """List of `~hips.HipsTile` (cached on multiple access)."""
+        tile_fetcher = HipsTileFetcher(self.tile_indices, self.draw_hips_order, self.hips_survey,
+                                       self.tile_format, self.progress_bar)
         if self._tiles is None:
-            self._tiles = list(self._fetch_tiles())
+            if self.use_aiohttp:
+                import asyncio
+                self._tiles = asyncio.get_event_loop().run_until_complete(tile_fetcher.tiles_aiohttp)
```
cdeil: The `import asyncio` shouldn't appear in `paint.py`, users of the fetcher shouldn't have to know anything about the implementation.
Please move that to the fetcher, and expose from the fetcher only a synchronous API, where the user requests a list of tiles, the call blocks, and is handed back a list of tiles (or error info about missing requests / tiles) when the call comes back.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,112 @@
...
+                 tile_format: str, progress_bar: bool) -> None:
+        self.tile_indices = tile_indices
+        self.hips_order = hips_order
+        self.hips_survey = hips_survey
+        self.tile_format = tile_format
+        self.progress_bar = progress_bar
+
+    def fetch_tile_threaded(self, url: str) -> Generator:
+        """Fetch a HiPS tile asynchronously."""
+        with urllib.request.urlopen(url, timeout=60) as conn:
```
cdeil: Timeout shouldn't be hard-coded here, but a config options for users of the fetcher class.
And default should be 10 sec like we discussed.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,112 @@
...
+                order=self.hips_order,
+                ipix=healpix_pixel_index,
+                frame=self.hips_survey.astropy_frame,
+                file_format=self.tile_format,
+            )
+            tile_urls.append(self.hips_survey.tile_url(tile_meta))
+            tile_metas.append(tile_meta)
+
+        raw_responses = []
+        with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
```
cdeil: You shouldn't hard-code `max_workers=5`  here, but somehow give users of this class the option to control this parameter.

cdeil:
Concerning your question about the current bug that tilemeta / tile isn't matched correctly: I don't know.

Please add a unit test that fails and address the other points mentinoed above, and then I'll have a look.

If you have an idea how to fix the issue, just go for it and then I'll review.

adl1995:
@cdeil I've made the updates, however, I still don't know how to enforce order when fetching tiles, so the tests fail.

Also, can you please explain how I could remove changes from `astropy_helpers`?

cdeil:
> Also, can you please explain how I could remove changes from astropy_helpers?

Here's one way to do it:
```
cd astropy_helpers
git checkout 14ca346  # This is the version hips master points to
cd ..
git status
git add -u
git commit -m 'Undo change to astropy_helpers'
```
Then to avoid introducing a messy git history where the `astropy_helpers` version jumps back and forward again, I'd suggest to use
```
git fetch origin   # might be called upstream for you, I don't know
git rebase -i origin/master
```
and to declare this new commit a fix-up commit on top of 05a130 which is the commit where you introduced the accidental change:
```
pick 6c12958 Introduce asynchronous fetching of HiPS tiles
pick 0eba105 Add option to fetch tiles asynchronously using urllib
pick 05a130f Add tqdm progress bar to urllib and aiohttp fetch methods
fixup 295434b Undo change to astropy_helpers
```
Finally, `git push -f` the branch to Github and check the list of commits and diff there. Should be all good now.

You're probably aware, but just in case: get into the habit of always using `git status` and `git diff` to review changes before committing, to spare yourself (and me) the troubles of having to fix up your git history later.

cdeil:
I would suggest that the `HipsTileFetcher` is given a Python list of `HipsTileMeta` of tiles to fetch.
That is simpler and more flexible compared to what you have now.
It then also means that you can create the list of `HipsTile` that you return more easily, because you already have the `meta`.
For the async code, can't you have a system that makes it easy to attach the right data to the right meta? Maybe one option could be that you pass the `meta` for each tile to the async function and that function hands back the complete `tile`, no?

hips/draw/paint.py
```diff
@@ -109,30 +112,18 @@ def projection(self, tile: HipsTile) -> ProjectiveTransform:
...
-            yield tile
-
     @property
     def tiles(self) -> List[HipsTile]:
         """List of `~hips.HipsTile` (cached on multiple access)."""
+        tile_fetcher = HipsTileFetcher(self.tile_indices, self.draw_hips_order, self.hips_survey,
+                                       self.tile_format, self.progress_bar)
         if self._tiles is None:
-            self._tiles = list(self._fetch_tiles())
+            if self.fetch_package == 'aiohttp':
```
cdeil: Here in `draw.py` you have the caller of the `HipsTileFetcher`. It should be presented with a uniform and simple interface for both cases. Suggest to pass the option how to fetch to `HipsTileFetcher.__init__` and then run and access the tiles in a uniform way here, without if / else. In practice this means that you move this if / else to a method on the `HipsTileFetcher`.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,120 @@
...
+    def fetch_tile_threaded(self, url: str, timeout: int = 10) -> Generator:
+        """Fetch a HiPS tile asynchronously."""
+        with urllib.request.urlopen(url, timeout=timeout) as conn:
+            return conn.read()
+
+    @property
+    def tiles(self) -> np.ndarray:
+        """Generator function to fetch HiPS tiles from a remote URL."""
+        tile_urls, tile_metas = [], []
+        for healpix_pixel_index in self.tile_indices:
```
cdeil: I see you have this code creating a list of HipsTileMeta twice. If you just take this list as input to the fetcher, then you can remove this code duplication.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,120 @@
...
+                frame=self.hips_survey.astropy_frame,
+                file_format=self.tile_format,
+            )
+            tile_urls.append(self.hips_survey.tile_url(tile_meta))
+            tile_metas.append(tile_meta)
+
+        tasks = []
+        async with aiohttp.ClientSession() as session:
+            for idx, url in enumerate(tile_urls):
+                task = asyncio.ensure_future(self.fetch_tile_threaded_aiohttp(url.format(idx), session))
```
cdeil: I see your `fetch_tiles_aiohttp` calls a method called `fetch_tile_threaded_aiohttp`.
async doesn't use threads, it runs in a single thread.
Maybe just rename the method to be accurate?

hips/tiles/fetch.py
```diff
@@ -0,0 +1,120 @@
...
+            if self.progress_bar:
+                from tqdm import tqdm
+                raw_responses = []
+                for f in tqdm(asyncio.as_completed(tasks), total=len(tasks), desc='Fetching tiles'):
+                    raw_responses.append(await f)
+            else:
+                raw_responses = await asyncio.gather(*tasks)
+
+        tiles = []
+        for idx, raw_data in enumerate(raw_responses):
```
cdeil: I see you have a for look using an `idx` before and after the parallel fetching.
I think if you follow my earlier suggestion and pass the `meta` to the function that does the fetch and makes the tile, then this `idx` isn't needed anymore?

hips/tiles/fetch.py
```diff
@@ -0,0 +1,120 @@
...
+        List of index values for HiPS tiles
+    hips_order : int
+        Order of the HiPS survey
+    hips_survey : `~hips.HipsSurveyProperties`
+        HiPS survey properties
+    tile_format : {'fits', 'jpg', 'png'}
+        Format of HiPS tile
+    progress_bar : bool
+        Show a progress bar for tile fetching and drawing
+    thread_count : int
```
cdeil: I see here in the `HipstileFetcher` you have an option `thread_count`. Please rename it to `n_parallel` or something that doesn't contain the word "thread", and then also use it to set the max number of parallel requests for the async fetching. I.e. this option should have an effect and work the same for both methods of fetching you implement.

Furthermore, the `timeout` should be listed here as well, and then again used in the same way for both methods to fetch.

adl1995:
@cdeil Your suggestion on coupling `HipsTileMeta` with the fetch method worked, thanks!

I think this is ready to be merged now. Handling the fetch errors and missing tiles could be a separate PR.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,113 @@
...
+        self.hips_survey = hips_survey
+        self.progress_bar = progress_bar
+        self.n_parallel = n_parallel
+        self.timeout = timeout
+        self.fetch_package = fetch_package
+
+    @property
+    def tile_urls(self) -> List[str]:
+        """List of tile URLs"""
+        tile_urls = []
```
cdeil: This can be a one-liner:
```python
return [self.hips_survey.tile_url(meta) for meta in self.tile_metas]
```

hips/tiles/fetch.py
```diff
@@ -0,0 +1,113 @@
...
+        for meta in self.tile_metas:
+            tile_urls.append(self.hips_survey.tile_url(meta))
+
+        return tile_urls
+
+    @property
+    def tiles(self):
```
cdeil: Missing return type annotation and docstring.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,113 @@
...
+        """Fetch a HiPS tile asynchronously."""
+        with urllib.request.urlopen(url, timeout=self.timeout) as conn:
+            raw_data = conn.read()
+            return HipsTile(meta, raw_data)
+
+    @property
+    def tiles_urllib(self) -> np.ndarray:
+        """Generator function to fetch HiPS tiles from a remote URL."""
+        with concurrent.futures.ThreadPoolExecutor(max_workers=self.n_parallel) as executor:
+            future_to_url = {executor.submit(self.fetch_tile_urllib, url, self.tile_metas[idx]) : url for idx, url in enumerate(self.tile_urls)}
```
cdeil: Please break the line to make it readable.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,113 @@
...
+
+            if self.progress_bar:
+                from tqdm import tqdm
+                requests = tqdm(concurrent.futures.as_completed(future_to_url), total=len(future_to_url), desc='Fetching tiles')
+            else:
+                requests = future_to_url#concurrent.futures.as_completed(future_to_url)
```
cdeil: What's up with the commented-out code?
Is this old / can be removed or is there a TODO here?

hips/tiles/fetch.py
```diff
@@ -0,0 +1,113 @@
...
+
+            tiles = []
+            for future in requests:
```
cdeil: I find this confusing that you have names "future" and "requests".
Please either choose "future" and "futures",
or choose "request" and "requests".

hips/tiles/fetch.py
```diff
@@ -0,0 +1,113 @@
...
+                for f in tqdm(tasks, total=len(tasks), desc='Fetching tiles'):
+                    tiles.append(await f)
+            else:
+                tiles = await asyncio.gather(*tasks)
+
+        return tiles
+
+    @property
+    def tiles_aiohttp(self) -> np.ndarray:
+        import asyncio
```
cdeil: `asyncio` is Python stdlib. Please move this import to the top of the file.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,113 @@
...
+        return tiles
+
+    async def fetch_tile_aiohttp(self, url: str, meta : HipsTileMeta, session) -> Generator:
+        """Fetch a HiPS tile asynchronously using aiohttp."""
+        async with session.get(url) as response:
+            raw_data = await response.read()
+            return HipsTile(meta, raw_data)
+
+    @property
+    async def fetch_all_tiles_aiohttp(self) -> np.ndarray:
```
cdeil: I think the return type annotation here is wrong, no?

hips/tiles/fetch.py
```diff
@@ -0,0 +1,113 @@
...
+            raise ValueError(f'Invalid package name: {self.fetch_package}')
+
+    def fetch_tile_urllib(self, url: str, meta : HipsTileMeta) -> Generator:
+        """Fetch a HiPS tile asynchronously."""
+        with urllib.request.urlopen(url, timeout=self.timeout) as conn:
+            raw_data = conn.read()
+            return HipsTile(meta, raw_data)
+
+    @property
+    def tiles_urllib(self) -> np.ndarray:
```
cdeil: Wrong return type annotation.

cdeil:
One more thing: I think it would be good to have at least two or three sentences of description in the fetcher saying what it does and at least one example how it works. Note that this is code-wise by far the most complex piece in our package (using threads and async and requests) and soon we will have to extend it with error handling in some way and I think it would be really valuable to have a short high-level description what it does and how it works for future developers.

Users will probably never use it directly. For them the options in `make_sky_image` matter. There you currently only expose the choice whether to use urllib or aiohttp, but not n_parallel or timeout. I think it would be especially important to expose `timeout`, because from experience I know that it can be super frustrating to have something time out if one a slow but working network, with no option to increase the timeout parameter. The options to expose it would be either individually, or as a dict "fetch_opts" where the user can pass e.g. `fetch_opts=dict(package='urllib', timeout=30, n_parallel=2)`

adl1995:
@cdeil I just made the updates. I have removed the `HipsTileFetcher` class and now the `fetch.py` file exports a `fetch_tiles` function. I'm not sure if this is neat, as now I had to pass around arguments that were previously fetched using `self`.

adl1995:
@cdeil I have refactored fetching tiles using `urllib`, now `tiles_urllib` function computes `url` on the fly and passes it to `executor.submit`. This works, but I have a feeling that tiles are not being fetched asynchronously, I say this because it took me 89 seconds for fetching tiles when building the docs.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,130 @@
...
+    n_parallel : int
+        Number of threads to use for fetching HiPS tiles
+    timeout : int
+        Seconds to timeout for fetching a HiPS tile
+    fetch_package : {'urllib', 'aiohttp'}
+        Package to use for fetching HiPS tiles
+
+    Examples
+    --------
+    >>> from hips import HipsSurveyProperties
```
cdeil: I tried the example you give, looks like it's broken:
```
In [1]: %paste
>>> from hips import HipsSurveyProperties
>>> from hips import fetch_tiles
>>> url='http://alasky.unistra.fr/DSS/DSS2Merged/properties',
>>> hips_survey = HipsSurveyProperties.fetch(url)
>>> tile_indices = [69623, 69627, 69628, 69629, 69630, 69631]
>>> tile_metas = []
>>> for healpix_pixel_index in tile_indices:
...    tile_meta = HipsTileMeta(
...        order=7,
...        ipix=healpix_pixel_index,
...        frame=hips_survey.astropy_frame,
...        file_format='fits',
...    )
...    tile_metas.append(tile_meta)
>>> tiles = fetch_tiles(tile_metas, hips_survey)
## -- End pasted text --

WARNING: version mismatch between CFITSIO header (v3.39) and linked library (v3.35).

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-cc871f4557f5> in <module>()
      2 from hips import fetch_tiles
      3 url='http://alasky.unistra.fr/DSS/DSS2Merged/properties',
----> 4 hips_survey = HipsSurveyProperties.fetch(url)
      5 tile_indices = [69623, 69627, 69628, 69629, 69630, 69631]
      6 tile_metas = []

~/code/hips/hips/tiles/survey.py in fetch(cls, url)
     86         """
     87 
---> 88         with urllib.request.urlopen(url) as response:
     89             text = response.read().decode('utf-8')
     90         return cls.parse(text, url)

/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in urlopen(url, data, timeout, cafile, capath, cadefault, context)
    221     else:
    222         opener = _opener
--> 223     return opener.open(url, data, timeout)
    224 
    225 def install_opener(opener):

/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py in open(self, fullurl, data, timeout)
    515                 req.data = data
    516 
--> 517         req.timeout = timeout
    518         protocol = req.type
    519 

AttributeError: 'tuple' object has no attribute 'timeout'
```

hips/tiles/fetch.py
```diff
@@ -0,0 +1,130 @@
...
+    """Generator function to fetch HiPS tiles from a remote URL."""
+    with concurrent.futures.ThreadPoolExecutor(max_workers=n_parallel) as executor:
+        if progress_bar:
+            from tqdm import tqdm
+            tile_metas = tqdm(tile_metas, total=len(tile_metas), desc='Fetching tiles')
+
+        tiles = []
+        for meta in tile_metas:
+            url = hips_survey.tile_url(meta)
+            future = executor.submit(
```
cdeil: I think this call fits on one line, no?

hips/tiles/fetch.py
```diff
@@ -0,0 +1,130 @@
...
+        raise ValueError(f'Invalid package name: {fetch_package}')
+
+def fetch_tile_urllib(url: str, meta: HipsTileMeta, timeout: float) -> Generator:
+    """Fetch a HiPS tile asynchronously."""
+    with urllib.request.urlopen(url, timeout=timeout) as conn:
+        raw_data = conn.read()
+        return HipsTile(meta, raw_data)
+
+def tiles_urllib(tile_metas: List[HipsTileMeta], hips_survey : HipsSurveyProperties,
+                 progress_bar: bool = False, n_parallel: int = 10,  timeout: int = 10) -> List[HipsTile]:
```
cdeil: Put timeout as "float" everywhere in type annotations and docstrings.

hips/tiles/fetch.py
```diff
@@ -0,0 +1,130 @@
...
+
+    return tiles
+
+async def fetch_tile_aiohttp(url: str, meta : HipsTileMeta, session) -> Generator:
+    """Fetch a HiPS tile asynchronously using aiohttp."""
+    async with session.get(url) as response:
+        raw_data = await response.read()
+        return HipsTile(meta, raw_data)
+
+async def fetch_all_tiles_aiohttp(tile_metas: List[HipsTileMeta], hips_survey : HipsSurveyProperties, progress_bar: bool) -> List[HipsTile]:
```
cdeil: Break the line

hips/tiles/fetch.py
```diff
@@ -0,0 +1,130 @@
...
+            for f in tqdm(tasks, total=len(tasks), desc='Fetching tiles'):
+                tiles.append(await f)
+        else:
+            tiles = await asyncio.gather(*tasks)
+
+    return tiles
+
+def tiles_aiohttp(tile_metas: List[HipsTileMeta], hips_survey : HipsSurveyProperties,
+                  progress_bar: bool) -> List[HipsTile]:
+    return asyncio.get_event_loop().run_until_complete(fetch_all_tiles_aiohttp(tile_metas, hips_survey, progress_bar))
```
cdeil: Break the line, e.g. like this:
```
return asyncio.get_event_loop().run_until_complete(
    fetch_all_tiles_aiohttp(tile_metas, hips_survey, progress_bar)
)
```

hips/tiles/fetch.py
```diff
@@ -0,0 +1,130 @@
...
+    'fetch_tiles',
+]
+
+__doctest_skip__ = [
+    'fetch_tiles',
+]
+
+
+def fetch_tiles(tile_metas: List[HipsTileMeta], hips_survey : HipsSurveyProperties,
+                progress_bar: bool = False, n_parallel: int = 10,  timeout: int = 10, fetch_package : str = 'urllib') -> List[HipsTile]:
```
cdeil: Break the line (and put timeout as float)

hips/tiles/fetch.py
```diff
@@ -0,0 +1,130 @@
...
+
+        tiles = []
+        for meta in tile_metas:
+            url = hips_survey.tile_url(meta)
+            future = executor.submit(
+                fetch_tile_urllib,
+                url,
+                meta,
+                timeout)
+            tiles.append(future.result())
```
cdeil: I think the problem why for threads it's executed synchronously is here.
You have to have one for loop that submits all the tasks, that will give you a list of futures.

And then only after everything is submitted you can call `.result()` or do some other blocking call that will wait for results to arrive.

adl1995:
@cdeil The example should be fixed now, there was a leading `,` after the `url`, so it caused errors. The tile fetching using `urllib` is now being done asynchronously.

cdeil:
@adl1995 - Thanks!

We could merge now if you like, but I have one remaining question / concern regarding how the async and thread parallel fetch is implemented
- in the thread solution, you manage to have the `if progress_bar` up front, but in the async done differently. Is there a way to make the thread and async implementation more similar or is that not possible because with async / await it's not possible to do it equivalently?
- another issue is that with the async solution, you have two different ways of awaiting the results, which means twice as much testing is needed. This is mostly the same comment, on whether it's possible to make the progress bar handling work the same for threads and async.
- and finally, a question for the thread solution: does it keep running 5 threads in the background, or does the number of parallel fetches go down because of the blocking `.result()` call, i.e. if a given tile is slow to fetch , but later ones are fast.

@adl1995 - If you're interested in fully understanding threads and async and what your code is doing (or maybe even writing another blog post), you could spend some time and use e.g. https://github.com/astrofrog/psrecord to "record' and plot or print a log of what is going on. I don't know if that can monitor number of threads and network I/O, but probably https://pythonhosted.org/psutil/ or some other tools can, and you could set of a "benchmark", i.e. a small Python file that executes a few fetches and / or draws and then makes a printout or plot summarising what is going on. This is something I wanted to do before implementing the faster tile drawing, but I won't have time for it. So if you're interested, take a day or two max for this "side project" and do that.

adl1995:
@cdeil There is an option to move the progress bar in the beginning in case of async, but it makes the code synchronous. The modified function is this:

```python
async def fetch_all_tiles_aiohttp(tile_metas: List[HipsTileMeta],
                                  hips_survey : HipsSurveyProperties, progress_bar: bool) -> List[HipsTile]:
    """Generator function to fetch HiPS tiles from a remote URL using aiohttp."""
    import aiohttp

    async with aiohttp.ClientSession() as session:
        if progress_bar:
            from tqdm import tqdm
            tile_metas = tqdm(tile_metas, total=len(tile_metas), desc='Fetching tiles')

        tiles = []
        for meta in tile_metas:
            url = hips_survey.tile_url(meta)
            task = asyncio.ensure_future(fetch_tile_aiohttp(url, meta, session))
            tiles.append(await task)
    return tiles
```

adl1995:
@cdeil I made the updates. However, there is a small problem. When using `asyncio.as_completed` the tiles are not fetched in sequence, this is no issue during drawing, but in `test_fetch.py` I assert on the first tile, which can be at a different index each time.

So, for now, I just fetched a single tile and asserted on that. I found this [question on SO](https://stackoverflow.com/questions/35699601/python-asyncio-as-completed-in-order), but did not try to implement this, because this would create an extra overhead which isn't exactly needed, except for the `test_fetch.py` file.

cdeil:
@adl1995 - I've done one commit 1fd6e7f to (hopefully) improve the code a little bit.
It's minor changes, mostly to avoid very long lines.
One change is that I managed to get rid of the no-op "progress_bar", by simply wrapping an iterable of futures with the `tqdm` call, again returning an iterable.
I think / hope it's still working as before.

---

One thing I noticed is that for the async, you haven't actually implemented the `n_parallel` and `timeout` options. Can you please add it?
Both are really important for the caller to have control over.

adl1995:
@cdeil Can you please tell how I could pull in your commit? I messed up my Git history.

cdeil:
The branch on Github is OK and contains my commit: https://github.com/hipspy/hips/pull/106/commits

You could just delete your local branch, and check out a new local branch from this latest remove version, and then continue coding there, no?

adl1995:
@cdeil I have added `timeout` and `n_parallel` option to `tiles_aiohttp` function. I'm not sure why you removed `asyncio.as_completed`, doesn't this mean that if one tile takes too long to fetch, others have to wait?

adl1995:
@cdeil The tests fail with:

> ImportError: /home/travis/miniconda/envs/test/lib/python3.6/lib-dynload/../../libz.so.1: version `ZLIB_1.2.9' not found (required by /home/travis/miniconda/envs/test/lib/python3.6/site-packages/matplotlib/../../.././libpng16.so.16)

I'm not sure what causes this.

cdeil:
> I have added timeout and n_parallel option to tiles_aiohttp function.

Thanks!

---

>  I'm not sure why you removed asyncio.as_completed, doesn't this mean that if one tile takes too long to fetch, others have to wait?

I don't think I did remove it, just moved it up a few lines, no?

---

> The tests fail ...

This looks like a Linux install issue of our dependencies, not like something related to this PR. I've restarted travis-ci to see if it was a fluke.

cdeil:
travis-ci passed, it was a fluke.

I've done a final review and some small improvements in b4805ad :
- Fixed some type annotations
- Change default of n_parallel to 5, because I don't think that a higher value improves total fetch time, and if so, then more parallel requests in flight just means higher chance for one of them to time out.
- Remove defaults for helper functions, to avoid duplicating the default values in different parts of the code (hard to keep consistent)
- Use `aiohttp.TcpConnector` because I think it's a bit simpler then the `asyncio.Semaphore` (one less level of code indentation, and more powerful (has other options to control requests that we might want to use in the future).

Merging this now, and then I'll make a issue with my thoughts how to improve `fetch.py` further for you @adl1995 in a follow-up PR.

PR #109, opened 2017-09-25
Change from healpy to astropy-healpix

cdeil:
This is a first quick attempt to change the `hips` package from `healpy` over to the to-be Astropy healpix package (http://healpix.readthedocs.io/). The main motivation for the change is to support Windows, another motivation is to have an easier-to-install dependency (or no dependency at all if this gets merged as `astropy.healpix` into the core at some point).

This pull request is work in progress, it should only be merged after there has been a release and a conda package exists for the Astropy healpix package.

In addition, I'd like to remove `hips/utils/healpix.py` and change the callers to use the main Astropy healpix API, not the healpy-compatible API, and the dependency declaration in `setup.py` as well as the install docs need to be updated.

@astrofrog - For now, I see two issues when running the tests:
https://gist.github.com/cdeil/a81421e94df6ed4cd732ddcf61c4174a

1. We are passing 2-dim arrays, but currently only 1-dim arrays are supported. This is issue https://github.com/cdeil/healpix/issues/21 . This is a blocker for making the change.
2. We are calling `healpy.boundaries` which is still missing in `healpix.healpy`. This is issue https://github.com/cdeil/healpix/issues/23 . This means it's not a drop-in replacement, but we should change to `healpix_boundaries_lonlat` anyways, so not really needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,5 @@ distribute-*.tar.gz
 
 # Mac OSX
 .DS_Store
+
+v
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,8 @@ env:
         - MAIN_CMD='python setup.py'
         - SETUP_CMD='test'
         - EVENT_TYPE='pull_request push'
-        - CONDA_DEPENDENCIES='healpy scikit-image Pillow reproject matplotlib tqdm aiohttp'
-        - PIP_DEPENDENCIES=''
+        - CONDA_DEPENDENCIES='astropy-healpix scikit-image Pillow reproject matplotlib tqdm aiohttp'
+#        - PIP_DEPENDENCIES='git+https://github.com/astropy/astropy-healpix.git#egg=astropy_healpix'
         - CONDA_CHANNELS='conda-forge astropy-ci-extras astropy'
         - SETUP_XVFB=True
         - DEBUG=True
@@ -56,7 +56,7 @@ matrix:
         # Check that install / tests work with minimal required dependencies
         - os: linux
           env: SETUP_CMD='test -V'
-               CONDA_DEPENDENCIES='healpy scikit-image Pillow'
+               CONDA_DEPENDENCIES='astropy-healpix scikit-image Pillow'
 
         # Try MacOS X
         - os: osx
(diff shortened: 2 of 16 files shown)


astrofrog:
> This is a first quick attempt to change the hips package from healpy over to the to-be Astropy healpix package (http://healpix.readthedocs.io/). The main motivation for the change is to support Windows, another motivation is to have an easier-to-install dependency (or no dependency at all if this gets merged as astropy.healpix into the core at some point).

Also licensing :)

@cdeil - ok, we should milestone cdeil/healpix#21 as 0.1. Do you have any time to work on this? Unfortunately I have to put work on healpix on hold for a few days while catching up with some other projects, so if you would have any time to work on it that would be great.

cdeil:
@astrofrog - I'm afraid I won't have time for https://github.com/cdeil/healpix/issues/21 this week. I'll take a quick stab at implementing https://github.com/cdeil/healpix/issues/31 (much simpler, smaller task) now, but probably won't even have time to finish that (i.e. add tests covering all edge cases, checking the maximum resolution we can support).

cdeil:
Locally now `hipspy` works with `astropy-healpix`. Let's see if we can make it work on the CI builds, especially Appveyor (Windows never worked until now, but now it should).

cdeil:
@astrofrog - Any idea why the Appveyor build tried to pip install scipy?
https://ci.appveyor.com/project/cdeil/hips-19275o3fsp6drby/build/1.0.336

I'll change Appveyor to Python 3.6 now, but I presume the scipy install fail will still be there.

cdeil:
It was because of scikit-image. Now CI should pass. I'll check back tomorrow.

cdeil:
First successful build / test on Windows: https://ci.appveyor.com/project/cdeil/hips-jra0zyeh9itvq3c/build/1.0.338

cdeil:
There is this one fail on Windows:
https://ci.appveyor.com/project/cdeil/hips-19275o3fsp6drby/build/1.0.340#L316

I think it's an int overflow. I'll try adding a `dtype=float` to the `np.sum` call - that should resolve the issue by using a float64 accumulator.

cdeil:
Yippie!

astrofrog:
Nice! 🎉

PR #114, opened 2017-10-28
Update readthedocs.yml to use astropy-healpix

cdeil:
This PR is updating readthedocs.yml to use astropy-healpix .

Really the main thing here is to see if the build / test works on Windows (see #3).


cdeil:
Ah, I forgot. Appveyor build was already working in #109 . Only the badge in README was pointing somewhere else. Fixed in 7681e6e. Merging this now and will cut a v0.2 release.

PR #123, opened 2018-07-04
Add healpix to hips conversion method

adonath:
This PR adds the healpix to hips conversion functionality. It is exposed via the method `healpix_to_hips`.
I added a simple test using a healpix image filled with `np.arange()` in the nested scheme and compared  the sum of the indices to manually created reference values. I also added an `environment.yml` defining a `hips-dev` environment.


hips/draw/healpix.py
```diff
@@ -0,0 +1,87 @@
...
+
+    for tile_idx in range(n_tiles):
+        tile = healpix_to_hips_tile(hpx_data=hpx_data, tile_width=tile_width,
+                                    tile_idx=tile_idx, file_format=file_format)
+
+        filename = Path(base_path) / tile.meta.tile_default_path
+        filename.parent.mkdir(exist_ok=True, parents=True)
+        tile.write(filename=filename)
+
+    # Write a minimal property file
```
cdeil: I think this class it a dict representation of the properties file?
https://hips.readthedocs.io/en/latest/api/hips.HipsSurveyProperties.html
Can you use it here and add the to_string and write there if it doesn't exit already?

Either way I'd suggest to use this for the write text:
https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text
It also closes the file on the end of the method call, and is one line less and doesn't need the with and local variable f, i.e. ist just a bit simpler overall.

adonath:
@cdeil OK to add the high level docs example later, once the support for color image is added?

cdeil:
Sure

cdeil:
@adonath - Thanks!

PR #124, opened 2018-07-05
Use dict (now ordered in Python 3.6+

cdeil:
Python dicts are ordered since Python 3.6, and it will stay that way (see [here](https://mail.python.org/pipermail/python-dev/2017-December/151283.html)).
So removing `collections.OrderedDict` by `dict`.


cdeil:
Looking at the logs, I think all of the fails are HTTP fetch timeouts, unrelated to this change. Merging.

Issue #127, opened 2018-07-14
support for overwrite in write_image


richardgmcmahon:
Hi, it would be useful if writeto supported 'overwrite' at least for fits files.

see:

astropy.io.fits.writeto(filename, data, header=None, output_verify='exception', overwrite=False, checksum=False

cdeil:
I would suggest we add an overwrite option here:
https://github.com/hipspy/hips/blob/3710d32caf1d3affc3e8f8046b7cc82b9cb71718/hips/draw/ui.py#L103

We could also add it here:
https://github.com/hipspy/hips/blob/8d6078ffb08f2fd1e63e54b305b5b336db8f535c/hips/tiles/tile.py#L340
although if we start to make it configurable when HiPS tile folders are created or files overwritten, things get more complicated, so I wouldn't add it there.

cdeil:
Implemented by @adl1995 in #129.

PR #129, opened 2018-07-20
Add overwrite option in write_image function for FITS files

adl1995:
Following issue #127, this PR adds the `overwrite` parameter to `write_image` function, which defaults to `False`. For now, this parameter only works with FITS files, I'm not sure what's the default behavior of PIL for JPG and PNG files.

@cdeil - Should I include a test case for this by writing a same file twice?

### DIFF
--- a/hips/draw/tests/test_ui.py
+++ b/hips/draw/tests/test_ui.py
@@ -74,12 +74,8 @@ def test_make_sky_image(tmpdir, pars):
     result.write_image(str(tmpdir / 'test.' + pars['file_format']), False)
 
     # Try and overwrite the image.
-    if pars['file_format'] == 'fits':
-        with pytest.raises(OSError):
-            result.write_image(str(tmpdir / 'test.' + pars['file_format']), False)
-    else:
-        with pytest.raises(FileExistsError):
-            result.write_image(str(tmpdir / 'test.' + pars['file_format']), False)
+    with pytest.raises(FileExistsError):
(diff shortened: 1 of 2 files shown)


cdeil:
@adl1995 - Thanks!

Well ideally, this option would work the same for FITS, PNG and JPEG.
I guess if Pillow doesn't offer the equivalent you have to do something like:
```
from pathlib import Path
if Path(filename).exits():
    raise ...
```

Concerning test: yes, would be nice and shouldn't be hard to write (using `with pytest.raises`), but I'd also merge without since the code is simple.

adl1995:
@cdeil - Please see the updated code.

For FITS files, [astropy.io.fits.writeto](http://docs.astropy.org/en/stable/io/fits/api/files.html#astropy.io.fits.writeto) raises the `OSError` exception by default. However, for JPG and PNG files I have manually raised the `FileExistsError` exception, as it seems more appropriate.

Is this fine, or should all formats raise the same exception for consistency?

cdeil:
I think if you move the code that checks and raises FileExistsError at the top, it will be nicer for the user, and already for you in the test, because it behaves the same.

Note that you should create an `FileExistsError` instance with the filename mentioned in the raise statement, as explained here:
https://stackoverflow.com/questions/36077266/how-do-i-raise-a-filenotfounderror-properly

adl1995:
This seems better. The `FileExistsError` is now raised for all file formats (FITS, JPG, PNG).

cdeil:
Looks good. Thanks!

Issue #130, opened 2018-07-21
how to make a hips plot without the tile boundaries plotted 


richardgmcmahon:
I am using the example plot_fits.py with a smaller fov:

https://github.com/hipspy/hips/blob/master/docs/getting_started.rst

width=360, height=360, fov="0.1 deg"

See attached png; How do I NOT get the tile boundaries plotted?

help(make_sky_image) does not tell me:

help(plot) tells me that the tiles arre are plotted but not how to NOT plot them
plot(self) -> None
 |      Plot the all sky image and overlay HiPS tile outlines.
```
result.plot()
# Draw the sky image
import matplotlib.pyplot as plt
from astropy.visualization.mpl_normalize import simple_norm
ax = plt.subplot(projection=geometry.wcs)
norm = simple_norm(result.image, 'sqrt', min_percent=1, max_percent=99)
ax.imshow(result.image, origin='lower', norm=norm, cmap='gray')

import matplotlib.pyplot as plt
plt.savefig('hips.png')
plt.show()
```
![hips](https://user-images.githubusercontent.com/2388403/43040905-0c22960c-8d49-11e8-9629-7199e6e305b9.png)

If I do not call result.plot() before the plt sequence I get:

![hips](https://user-images.githubusercontent.com/2388403/43040945-561ae2b8-8d4a-11e8-9af1-13aace57e89e.png)

So somewhere behind the scenes some matplotlib hidden magic is taking place. e.g. is make_sky_image doing some matplotlib work

As well as turning of the tile boundaries, it could be useful to know how to specify line color and line style for the tile boundaries too.

adl1995:
@cdeil - Would adding a section on "Plot using Astropy visualization toolkit" in the "Getting started" page be enough? We could redirect users to Astropy docs for further reading.

What sort of example should we provide using the `astropy.visualization` package? Currently, the example in the docs has this snippet:
```python
from astropy.visualization.mpl_normalize import simple_norm
ax = plt.subplot(projection=geometry.wcs)
norm = simple_norm(result.image, 'sqrt', min_percent=1, max_percent=99)
ax.imshow(result.image, origin='lower', norm=norm, cmap='gray')
```

Which produces this plot:
![astropy_plot](https://user-images.githubusercontent.com/13198325/43075478-713651bc-8e9a-11e8-8a94-3b7066eca2c3.png)

cdeil:
OK, if you want to keep the current single high-level docs page approach for now.

[...]

PR #131, opened 2018-07-24
Provide example on how to plot image using Astropy visualization toolkit

adl1995:
This PR updates the docs and adds a section named "Plot using Astropy visualization toolkit" in the "Getting started" page.

The [result.plot()](https://github.com/hipspy/hips/blob/master/hips/draw/ui.py#L124) method is also updated as it now uses a boolean `show_grid` to turn off the tile boundaries grid (Cf. #130).


cdeil:
@adl1995 - Thanks!

I don't have time this week for proper review, but from a very quick look this looks great, and tests pass, so I'll just merge this in.

Issue #120, opened 2018-04-21
Generate HEALPix map from HiPS


tboch:
Given a HiPS, it's quite easy to generate a HEALPix map, using the low-level tiles.
The needed code is already present in this gist:
https://gist.github.com/tboch/f68cd1bb1529d8ac12184b40e54ba692

PR #132, opened 2018-07-28
Add function for HiPS to HEALPix conversion 

adl1995:
Following issue #120, this pull request adds a function for generating a HEALPix map from HiPS.

The function is replicated from this GitHub Gist by @tboch: https://gist.github.com/tboch/f68cd1bb1529d8ac12184b40e54ba692

@cdeil - Can you please suggest some test cases for this?

### DIFF
--- a/hips/draw/healpix.py
+++ b/hips/draw/healpix.py
@@ -2,6 +2,7 @@
 import numpy as np
 import healpy as hp
 from pathlib import Path
+from typing import Union
 from astropy.io import fits
 
 from ..tiles import HipsTile, HipsTileMeta, HipsSurveyProperties
@@ -15,7 +16,8 @@ __all__ = [
 ]
 
 
-def healpix_to_hips_tile(hpx_data, tile_width, tile_idx, file_format) -> HipsTile:
+def healpix_to_hips_tile(hpx_data: np.ndarray, tile_width: int,
+                         tile_idx: int, file_format: str) -> HipsTile:
     """Create single hips tile from healpix data given a tile index.
 
     Parameters
@@ -59,7 +61,8 @@ def healpix_to_hips_tile(hpx_data, tile_width, tile_idx, file_format) -> HipsTil
     return HipsTile.from_numpy(meta=meta, data=data)
 
 
-def healpix_to_hips(hpx_data, tile_width, base_path, file_format='fits'):
+def healpix_to_hips(hpx_data: np.ndarray, tile_width: int,
+                    base_path: Union[str, Path], file_format='fits') -> None:
     """Convert HEALPix image to HiPS.
 
     Parameters
--- a/hips/draw/healpix.py
+++ b/hips/draw/healpix.py
@@ -1,9 +1,10 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import numpy as np
-import healpy as hp
 from pathlib import Path
-from typing import Union
+from typing import Union, List
 from astropy.io import fits
+from astropy_healpix import healpy as hp
+import healpy as hp
 
 from ..tiles import HipsTile, HipsTileMeta, HipsSurveyProperties
 from ..utils.healpix import hips_tile_healpix_ipix_array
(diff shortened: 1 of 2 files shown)


cdeil:
@adonath - Did you start coding up this functionality already?

@adl1995 - I don't think @adonath did, but I'm not sure.

If you want to go ahead with this, you will have to more closely study the implementation and test case at #123 for the direction HEALPix -> HiPS, and then do the other direction HiPS -> HEALPix as similar as possible.

My understanding is that the `hips_to_healpix_array` you introduce here basically does the same thing as the existing `hips_tile_healpix_ipix_array`
https://github.com/hipspy/hips/blob/0b52667f09fde5ea3d323b2ef7a6c2b0e064c2c8/hips/utils/healpix.py#L125
and you should use that one.

For the test case, one thing you could do is start with HEALPix, convert to HipS tiles (no need to write to disk) and convert back to HEALPix. The input and output should be identical.

hips/draw/healpix.py
```diff
@@ -88,3 +95,115 @@ def healpix_to_hips(hpx_data, tile_width, base_path, file_format='fits'):
...
+    healpix_to_xy = [0 for i in range(nsize ** 2)]
+
+    fill_array(xy_to_healpix, nsize, None)
+
+    for i in range(len(xy_to_healpix)):
+        healpix_to_xy[xy_to_healpix[i]] = i
+
+    return healpix_to_xy, xy_to_healpix
+
+def hips_to_healpix(hips_url: str, npix: int, hpx_output_path: str):
```
cdeil: Let's not mix computation with I/O.

I would suggest to just take the HEALPix pixels as a Numpy array:
https://hips.readthedocs.io/en/latest/api/hips.healpix_to_hips.html

This assumes that the HEALPix fits in memory, but that's the common use case, and supporting larger images requires fancy file I/O where only chunks of the image are memmapped. Let's not do that.

hips/draw/healpix.py
```diff
@@ -88,3 +95,115 @@ def healpix_to_hips(hpx_data, tile_width, base_path, file_format='fits'):
...
+
+    hips_tile_width = hips_properties.tile_width
+    order = int(np.log2(hips_tile_width))
+
+    healpix_to_xy, xy_to_healpix = hips_to_healpix_array(order)
+
+    healpix_map = np.empty(hp.nside2npix(2 ** 12), dtype=np.double) # 2 ** 12?
+    healpix_map[:] = hp.pixelfunc.UNSEEN
+
+    for ipix in range(npix):
```
cdeil: You should not have a double for loop.

The function should take a list (or iterable) of HiPSTile as input, and only have a loop over that.
The way you pre-allocate the output array before the loop loops OK.

There is no second loop over pixels within a tile, because the operations for those pixels are vectorised as Numpy expressions:
https://github.com/hipspy/hips/blob/0b52667f09fde5ea3d323b2ef7a6c2b0e064c2c8/hips/utils/healpix.py#L125

hips/draw/healpix.py
```diff
@@ -88,3 +95,115 @@ def healpix_to_hips(hpx_data, tile_width, base_path, file_format='fits'):
...
+        try:
+            tile = fits.open(f'{hips_url}/Norder3/Dir0/Npix{ipix}.fits')
+        except:
+            continue
+
+        # BSCALE and BZERO.
```
cdeil: You should not be concerned with BSCALE or BZERO here. A HipsTile has a property that gives you the data as a Numpy array.

hips/draw/tests/test_healpix.py
```diff
@@ -36,3 +37,10 @@ def test_healpix_to_hips(tmpdir, file_format):
 
     properties = (tmpdir / 'properties').read_text(encoding=None)
     assert file_format in properties
+
+def test_hips_to_healpix(tmpdir):
```
cdeil: I would prefer a simple test that doesn't use real data for two reasons:

- Hitting the network and disk is slow, and for the network error-prone
- Your test should end with an assert value on one or a few pixel values; if you take real data, it's not clear from looking at the test if the code works, you get an arbitrary value. If you set up a test case where you see what goes in from reading the test code, it's clear what should come out.

cdeil:
@adl1995 - I think @adonath only wanted to extend the HEALPix -> HipS direction to also work for PNG and JPEG, so most likely this PR doesn't overlap with that and you can just go ahead if you have time.

cdeil:
I'll start a new branch to develop this functionality.

Issue #112, opened 2017-10-16
ValueError from flipup call in Tile.to_numpy


cdeil:
Sometimes an error occurs in the `np.flipup` call here when converting JPEG to numpy array:
https://github.com/hipspy/hips/blob/ed7d02d979a5e4966efad509184e942ae487b160/hips/tiles/tile.py#L302
```python
    with Image.open(bio) as image:
                data = np.array(image)
                # Flip tile to be consistent with FITS orientation
                data = np.flipud(data)
```

> ValueError: Input must be >= 1-d.

I saw this here: https://travis-ci.org/hipspy/hips/jobs/288318105#L1611
and also locally (from a different test) here: https://gist.github.com/cdeil/ad3dd0dc6a6c1fe8d58ec22ecdad03d0

It doesn't seem to be deterministic, I re-ran the same test and the fail doesn't show up.

I'm not sure what causes it, it looks like Pillow doesn't read the file properly, but only sometimes!???
For now, I'll just keep this issue open as a reminder, and try one thing: move the `np.flipup` call out of the `with` block, to make sure that the file was closed and Pillow was done for sure:
```
            with Image.open(bio) as image:
                data = np.array(image)
                # Flip tile to be consistent with FITS orientation
            data = np.flipud(data)
```

cdeil:
I don't have time to look into this issue now, but wanted to leave a note:

When @adonath and @tboch implemented the HiPS gen, they noticed an issue that FITS I/O doesn't work properly for a Numpy array that's transposed, as the result of calling `np.rot90` is:
https://github.com/hipspy/hips/pull/123/files#diff-4de2f89531f836afdfd881e73aea8632R43
Probably here where we call `np.flipup` also a transposed Numpy array is created and from the comments above it seems that also Pillow I/O sometimes doesn't work properly with that.

There is  https://github.com/astropy/astropy/issues/2150 describing the issue for Gzipped FITS, but I'm not aware of an open issue in the Astropy tracker without Gzip compression, which I think is what we're using for HiPS.

So basically this issue needs someone to look in detail at how FITS / PNG / JPEG I/O react to passing in transposed Numpy arrays (add test cases for that), and where necessary add a `np.copy` or `np.ascontiguousarray` call to make sure a "normal" Numpy array is used in those places.
That should be done in the I/O code, not in random code places like here:
https://github.com/hipspy/hips/pull/123/files#diff-4de2f89531f836afdfd881e73aea8632R43

[...]

PR #140, opened 2019-01-08
Add tests for HiPS tile <-> Numpy conversions

cdeil:
Today I wanted to make the `healpix_to_hips` work for RGB JPEG and RGBA PNG data. I'll make a pull request for that shortly.

However, I noticed that there seems to be a bug in our `HipsTile.from_numpy` or `HipsTile.to_numpy` for the RGB JPEG case. This PR shows the issue via a regression test. It works for FITS and also for RGBA PNG as expected.

Looking for old issues I saw #112, possibly this is the same issue. I did try adding `.copy` calls before and after the `flipud` calls in `from_numpy`, no change, so transposed Numpy arrays don't seem to be the issue.

@tboch and maybe if you have time @adl1995 or @adonath - If you have time, could you please have a look at this and comment if you have any idea?

This is blocking the work on HEALPix -> HiPS and the v0.3 release, so we should try to figure this out ASAP.

### DIFF
--- a/hips/tiles/tests/test_tile.py
+++ b/hips/tiles/tests/test_tile.py
@@ -2,6 +2,7 @@
 from pathlib import Path
 import pytest
 from astropy.tests.helper import remote_data
+import numpy as np
 from numpy.testing import assert_allclose, assert_equal
 from ...utils.testing import get_hips_extra_file, requires_hips_extra
 from ..tile import HipsTileMeta, HipsTile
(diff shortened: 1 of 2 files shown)


cdeil:
Actually it looks like the issue is just "JPEG encoding noise", if I change the example to larger integer numbers in the `unit8` range (e.g. 0, 100, 200 instead of 0, 1, 2) they approximately round-trip.

https://nbviewer.jupyter.org/gist/cdeil/c5401336d7d5eb477263fb8b7b358de1

adl1995:
@cdeil - Thanks for looking into this.

Could this possibly have something to do with JPEG being a lossy format? We currently save the image using the [Image.save](https://pillow.readthedocs.io/en/3.1.x/reference/Image.html#PIL.Image.Image.save) function from Pillow, without providing any extra parameters for controlling the image quality.

This [documentation](https://pillow.readthedocs.io/en/3.1.x/handbook/image-file-formats.html#jpeg) for JPEG file format lists other keyword parameters that can be used e.g. `dpi` and `quality` (which is a number from 0 to 100, default is set to 75).

Maybe you could try setting this to 100 and see if it affects the output?

cdeil:
This is not changing any behaviour. It's just adding tests that prove that tile Numpy array conversions are working OK and doing some (accidentally applied) code formatting.

Issue #136, opened 2018-12-13
Support RGB in healpix -> hips


cdeil:
@adonath  added two functions for healpix -> hips in #123

We should extend those to also be able to work with RGB tiles, that's what we need for our main use case to make HiPS from our own HEALPix Fermi-LAT maps.

I think it should work OK to extend the existing functions to work for that case as well (like we do in most other places in hipspy), instead of adding new ones.

Exact API needs a bit of thought / discussion.

@adl1995 started to add the other direction in #132, so that's a bit related, but a separate task.

@adonath - Do you have three example RGB HEALPix maps from Fermi-LAT that we could use for a docs example? (ideally 1 MB or 10 MB, not something huge, so that it runs fast)
Alternatively we could look e.g. for Planck allsky maps in 3 bands in HEALPix and try to make a Hips from that.

cdeil:
Done in #141

PR #141, opened 2019-01-08
Support RGB in HEALpix -> HiPS

cdeil:
This PR addresses #136 and adds RGB image support for the HEALPix -> HiPS conversion.

I'll try to add a docs example that shows that it really works before merging this in.

### DIFF
--- a/hips/draw/healpix.py
+++ b/hips/draw/healpix.py
@@ -12,7 +12,11 @@ log = logging.getLogger(__name__)
 
 
 def healpix_to_hips_tile(
-    hpx_data: np.ndarray, tile_width: int, tile_idx: int, file_format: str, frame: str
+    hpx_data: np.ndarray,
+    tile_width: int,
+    tile_idx: int,
+    file_format: str,
+    frame: str = "icrs",
 ) -> HipsTile:
     """Create single HiPS tile from HEALPix data.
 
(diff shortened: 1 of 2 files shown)


cdeil:
@tboch - Could you please have a look at the docs example here?

Is it possible to show HiPS in Aladin Lite without Allsky.fits using some options?
Or do I have to generate it to make this work?

For Aladin Desktop I also don't manage to load the HiPS, if I just pass the folder name.

Probably there's still bugs and missing information ... ?

tboch:
@cdeil: currently, this can not be displayed in Aladin Lite for 2 reasons : AL currently ony supports JPEG/PNG tiles and also requires the presence of Allsky (work in progress to get rid of the latter limitation).

I did not manage either to visualize the generated HiPS in Aladin Desktop, I'll check with Pierre tomorrow, I'm sure there must be a way.

cdeil:
I'll change the example to PNG and also generate an AllSky. Then it should work in Aladin Lite.
The JS code to load the files from local disk is OK?

tboch:
> The JS code to load the files from local disk is OK?

almost. It should be:

```javascript
aladin.setImageSurvey(
        aladin.createImageSurvey(
            'test123', 'test123', '.', 'galactic', 1, {imgFormat: 'png'}
        )
    );
```

cdeil:
> AL currently ony supports JPEG/PNG tiles and also requires the presence of Allsky (work in progress to get rid of the latter limitation).

That would be nice to make it work without Allsky. Also, if possible, making it work for any order, not just >=3 would be nice. I'll try to change to order=3 for the test and see if that works.

cdeil:
I've continued with this a bit in 866748ac20b0ab91c7829a34b37e3b288124eba5 .

I think it would be better to have the `order` as parameter from the start and to pass it in to all the functions, instead of inferring it from the HEALPix array size and tile width deep down in `healpix_to_hips_tile`:
https://github.com/hipspy/hips/pull/141/files#diff-4de2f89531f836afdfd881e73aea8632R78
I'll try that now.

cdeil:
With 6df49aa we now have for the first time generated some HiPS that Aladin Lite can load:

screenshot 2019-01-10 at 17 47 46


However the tiles clearly aren't filled correctly yet. Could be a bug in many places, e.g. the allsky tile handling or the hipsgen part. @tboch - If you have time to have a look, please do. Any suggestions / code review highly welcome. Otherwise I'll try to come back to this and figure it out / finish it up in the coming days.

Issue #79, opened 2017-07-19
Improve draw algorithm to handle tile edges in a good way


cdeil:
As discussed previously, our current method of combining projected tiles isn't good. We simply make full sky images for each projected tile and sum them up. This often leads to "bright lines" at tile edges, because pixels there are drawn and summed twice, i.e. the intensity is a factor 2 too high.

Example:  https://gist.github.com/cdeil/796875efe4a3620987b9f62eb278cfbc
Output:
![download](https://user-images.githubusercontent.com/852409/28357762-ca2f59ea-6c6c-11e7-900f-55c38e3b49eb.png)

---

Using the RGB example from the high-level docs, one can see that it can also happen that we get "dark lines" or holes between tiles when drawing with the current method:

![download2](https://user-images.githubusercontent.com/852409/28357841-f83ed554-6c6c-11e7-8045-712181bd6793.png)

---

And the current `SimpleTilePainter` is super slow, presumably because it applies the projective transform for every tile for the whole sky image (even though the tile only contributes to a small subset of pixels in the output image).

---

So we need to implement a better method in the `SimpleTilePainter`, to avoid these artifacts and get reasonable performance. I have some ideas, @adl1995 - we can discuss in the call this afternoon.

cdeil:
I'm moving this to the 0.2 milestone. I didn't have time to implement this before going on vacation and want to release v0.1 now.

cdeil:
Coming back to this question how to draw tiles.

The core of the current implementation is here:

https://github.com/hipspy/hips/blob/8d6078ffb08f2fd1e63e54b305b5b336db8f535c/hips/draw/paint.py#L199-L203

So the algorithm is to project each tile into a full sky image, and to sum up those projected sky images.
That is very inefficient in terms of storage space and CPU time.
And using `+` is also incorrect, it leads to the "bright edges" where pixel values are double of what they should be, as shown in the example above.

I think a much better algorithm would be to first compute a "plan" for a given sky image, where for every pixel in the sky image a decision is made which tile to use to compute it's value. This will give you a label image which you can integer-encode e.g. like this:
```
0 0 1 1
0 1 1 1
0 0 1 2
```
where `0` means use tile 0 and `1` means use tile `1` and `2` means use tile 2.

Then do something like this:
```
for tile_idx in range(0, label_image.max()):
    pix_idx = np.nonzero(image == tile_idx)
    # project and fill just these pixels for that tile
```
[...]

PR #144, opened 2019-08-10
WIP: Compute image ipix and only update masked indexes in all sky image

adl1995:
This is a followup of #79. Previously, the output image contained bright lines as the values were being summed twice at tile edges. The current change applies the resulting tile image only at certain indices, which it receives by applying a mask on the all sky image based on the ipix values.

Previous output:
result_unmasked

Current output:
result_masked

However, the updated output image now contains 'holes' where pixels values are not inserted. I believe this is an issue with the warping function we're currently using. I will trying using either  `skimage.transform.warp_coords` or `scipy.ndimage.map_coordiantes` and see if it resolves the issue.

### DIFF
--- a/hips/draw/paint.py
+++ b/hips/draw/paint.py
@@ -3,6 +3,7 @@ import time
 import numpy as np
 from typing import List, Tuple, Union, Dict, Any
 from astropy.wcs.utils import proj_plane_pixel_scales
+from astropy_healpix.high_level import HEALPix
 from skimage.transform import ProjectiveTransform, warp
 from ..tiles import HipsSurveyProperties, HipsTile, HipsTileMeta, fetch_tiles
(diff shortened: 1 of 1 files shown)


cdeil:
@adl1995 - Thanks for attempting to improve this!

To avoid issues at the edges, you have to call a function in scipy.ndimage or scikit-image to compute the pixel values at "target" coordinates that has an option or even by default extrapolates, i.e. if the target coordinate is just slightly outside the tile image, it will fill the tile edge pixel value and not black or white.

And to be efficient, you have to first compute the target coordinate array, only for the pixels for which a given tile should be used, and only then call the "reproject" or "interpolate" function to compute the pixel values for exactly those target positions.

Hope it's clear ...

PR #146, opened 2020-03-06
Allow overwriting of output files

bryanmiller:
Add the overwrite parameter so that output files can be overwritten.

### DIFF
--- a/hips/draw/ui.py
+++ b/hips/draw/ui.py
@@ -116,10 +116,10 @@ class HipsDrawResult:
 
         if self.tile_format == 'fits':
             hdu = fits.PrimaryHDU(data=self.image, header=self.geometry.fits_header)
-            hdu.writeto(filename)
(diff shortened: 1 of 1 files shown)


hips/draw/ui.py
```diff
@@ -116,10 +116,10 @@ def write_image(self, filename: str, overwrite: bool = False) -> None:
 
         if self.tile_format == 'fits':
             hdu = fits.PrimaryHDU(data=self.image, header=self.geometry.fits_header)
-            hdu.writeto(filename)
+            hdu.writeto(filename, overwrite=overwrite)
         else:
             image = Image.fromarray(self.image)
-            image.save(filename)
+            image.save(filename,j, overwrite=overwrite)
```
adl1995:
```suggestion
            image.save(filename, overwrite=overwrite)
```
I think you added the extra `j` by mistake. Otherwise, this looks good to me.

PR #154, opened 2023-02-05
Allow overwriting FITS images and fixing bright pixels in output

smeingast:
Let the writeto method for FITS images check for existing files and allow overwriting. Only check for existing file if image is to be saved via PIL.


smeingast:
OK, I accidentally added another modification to this branch where I modified the output of the draw_all_tiles method to avoid having bright pixels from summing. This is done by first creating an empty image (NaNs instead of 0s) and then iteratively stacking them with all individual tiles

